PR #33, opened 2023-01-15
Self installation and UI

Qubus0:
lots of stuff in here. 
the big stuff to test: 
1. automatic loader setup: add this startup flag `--script addons/mod_loader_tools/mod_loader_setup_helper.gd` and nothing else. Start the game from the editor or from steam. I won't tell you what the things in the UI are. if something is unclear, tell me because then it needs to be changed.
   1. try out the method mentioned in the readme under `For Mod Loader Developers > Clean Setup`
2. Redoing the loader setup via button and restarting
3. Toggling mods on and off (use these two small ones that only print stuff for you to test with: [download mods.zip](https://github.com/GodotModding/godot-mod-loader/files/10421417/mods.zip)). A disabled mod should not be initialised upon game start.
4. skipping the selection screen (via flag, pressing <kbd>ALT</kbd> or via the option in the selection)

Stuff that's not done yet:
1. mod icons
2. flag for autoload mods that are always loaded right away (state can only be changed when the game is restarted)

Let me know what you think!


ithinkandicode:
Compiled PCK for Brotato, hosted this myself due to GitHub's upload limits:

- https://brotato.codemuffin.com/_files/Brotato--ModLoader-PR-33.zip

Two test mods, they simply append the version on the title screen. 
The 2nd has the 1st as a dependency, so it won't work without it:

- [Darkly77-VersionString1--3.1.1.zip](https://github.com/GodotModding/godot-mod-loader/files/10424233/Darkly77-VersionString1--3.1.1.zip)
- [Darkly77-VersionString2--3.1.1.zip](https://github.com/GodotModding/godot-mod-loader/files/10424234/Darkly77-VersionString2--3.1.1.zip)

ithinkandicode:
**Screenshots**

Initial Setup - always shown on first load:
PR33-1

After initial setup, with no mods in `./mods`. Starts in fullscreen:
PR33-2

After adding mods and restarting:
PR33-3

After disabling the fullscreen through Brotato's own options, restarting, and enabling 1 mod:
PR33-4

README.md
```diff
@@ -2,9 +2,137 @@
 
 A general purpose mod-loader for GDScript-based Godot Games.
```
ithinkandicode, line 4: I don't think we should remove the link to the Delta-V Modding repo until our docs already cover everything they've said. Atm their repo has a lot of important and useful info that our readme doesn't have yet

README.md
```diff
@@ -2,9 +2,137 @@
...
 
-For detailed info, see the [docs for Delta-V Modding](https://gitlab.com/Delta-V-Modding/Mods/-/blob/main/MODDING.md), upon which ModLoader is based. The docs there cover mod setup and helper functions in much greater detail.
-
-## Mod Setup
+## Table of Contents
```
ithinkandicode, line 5: I don't think we should remove the link to the Delta-V Modding repo until our docs already cover everything they've said. Atm their repo has a lot of important and useful info that our readme doesn't have yet

Qubus0: i've just moved it further down https://github.com/GodotModding/godot-mod-loader/pull/33/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R206

README.md
```diff
@@ -2,9 +2,137 @@
...
+      * [appendNodeInScene](#append_node_in_scene)
+      * [saveScene](#save_scene)
+  * [For Game Developers](#for-game-developers)
+    * [How to integrate the ModLoader into a game](#how-to-integrate-the-modloader-into-a-game)
+  * [For Mod Loader Developers](#for-mod-loader-developers)
+    * [Clean Setup](#clean-setup)
+  * [For Everyone](#for-everyone)
+    * [Folder locations](#folder-locations)
+  * [Credits](#credits)
+<!-- TOC -->
```
ithinkandicode, line 27: We discussed this on Discord but I'll just reiterate it here for posterity: It would be good to move the bulkier parts of the README to the Wiki, and just have the README cover the basics -- ie. you land on the readme page and it tells you everything you need to know about what ModLoader is, the basics of the setup, what the options are -- with Wiki links for in-depth info

Qubus0:
I will decouple the interface from the self setup and the loader so the user can choose if they want to use it or an external tool/ do it manually

README.md
Qubus0: yes, completely agree. it's just something for another pr

README.md
```diff
@@ -2,9 +2,137 @@
...
+
+The mod loader comes with a little helper script to properly install itself without having to recompile a game's source.
+This works by using Godot's functionality to override project settings with a file named `override.cfg`.
+Since the override also saves all other Project settings, it's important to recreate this file after the game is updated (This can be done with one button press in the mod selection screen).
+
+Use this flag to skip the mod selection screen (or tick the box in the selection screen)
+`--skip-mod-selection`
+
+Holding <kbd>ALT</kbd> during startup will invert the behavior: 
+skip selection screen without the skip flag and show it with the skip flag
```
ithinkandicode, line 54: When you launch the GUI there's an option to disable the mod selection screen and only enable it if you're holding ALT. I didn't want to use that option because I don't know how to change it later, so it would be good to tell users how to do that

README.md
```diff
@@ -2,9 +2,137 @@
...
+   - something that's not an asset 
+   - something that's not game code (not .gd)
+3. if unsure, better ask the community!
+
+Sources to download from are -
+Steam workshop, for officially supported Steam games
+Thunderstore, for unsupported and non-Steam games
+Official Discord communities (be careful)
+
+Have a mod? Good. Don't unzip it and place it inside the `mods` folder alongside the game (see [Folder locations](#folder-locations))
```
ithinkandicode, line 72: The prior info is good but superfluous if you just want to know how to install a mod. The extra info can go on the Wiki

README.md
```diff
@@ -2,9 +2,137 @@
...
+
+## For Mod Developers
+
+### Development Setup
+
+**Prerequisites**
```
ithinkandicode, line 78: Why use bold instead of subheadings? Seems like a downgrade, as visually they look the same, but using headings lets you link to sections

README.md
```diff
@@ -2,9 +2,137 @@
...
+
+Use this flag to skip the mod selection screen (or tick the box in the selection screen)
+`--skip-mod-selection`
+
+Holding <kbd>ALT</kbd> during startup will invert the behavior: 
+skip selection screen without the skip flag and show it with the skip flag
+
+### Installing mods
+First, you will have to download a mod for your game. When downloading a mod always keep the basics in mind: 
+1. always confirm your sources are trustworthy
```
ithinkandicode, line 58: We should use proper grammar in the README

README.md
```diff
@@ -2,9 +2,137 @@
...
+   - random executables
+   - a mod using the OS class (has access to your computer system)
+   - something that's not an asset 
+   - something that's not game code (not .gd)
+3. if unsure, better ask the community!
+
+Sources to download from are -
+Steam workshop, for officially supported Steam games
+Thunderstore, for unsupported and non-Steam games
+Official Discord communities (be careful)
```
ithinkandicode, line 70: Might be confusing as users may expect this to mean there's some kind of Workshop/Thunderstore integration built in, but there isn't yet

README.md
```diff
@@ -2,9 +2,137 @@
...
+> `Note:` do not share any of these files unless you get explicit permission from the developer
+
+**Running the game from the Godot editor**
+
+When opening the project for the first time: 
+- select `import` (on the right) and navigate to the destination folder you previously selected. 
+Select the `project.godot` file and press `open and edit`.
+  - To run the game, press the Play triangle button in the top right.
+
+> `Note for Steam Games:`
```
ithinkandicode, line 100: Why use quote here? Might be better using details:

```
<details>
<summary><b>Name</b></summary>
Collapsed content
</details>
```

Which results in:

<details>
<summary><b>Name</b></summary>
Collapsed content
</details>

README.md
```diff
@@ -2,9 +2,137 @@
...
+
+### Developing a Mod
+
+First off, if the game you want to mod does not have this ModLoader installed already, consider bringing it up to the developer for them to consider. 
+Second off, there are some limitations if that's the case. There is no way to modify autoload (singleton) behavior, since the load order of 
+them can't be overridden due to some strange behavior in Godot's way of overriding them.
+
+#### Where to start?
+
+Download the starterkit mod fitting your game from this repository [Starter Kit Mods]()
```
ithinkandicode, line 130: missing URL

README.md
```diff
@@ -19,21 +147,25 @@ yourmod.zip
         └───manifest.json
 ```
 
-#### Notes on .import
-
-Adding the .import directory is only needed when your mod adds content such as PNGs and sound files. In these cases, your mod's .import folder should **only** include your custom assets, and should not include any vanilla files.
-
-You can copy your custom assets from your project's .import directory. They can be easily identified by sorting by date. To clean up unused files, it's helpful to delete everything in .import that's not vanilla, then run the game again, which will re-create only the files that are actually used.
+> `Notes on .import`
```
ithinkandicode, line 150: see comment re: quotes vs. details

addons/mod_loader_tools/mod_loader_helper.gd
```diff
@@ -0,0 +1,123 @@
...
+	return cmd_line_arg_exists("--skip-mod-selection") or\
+		is_project_setting_true(settings.SKIP_MOD_SELECTION)
+
+
+static func is_project_setting_true(project_setting: String) -> bool:
+	return ProjectSettings.has_setting(project_setting) and\
+		ProjectSettings.get_setting(project_setting)
+
+
+static func cmd_line_arg_exists(argument: String) -> bool:
```
ithinkandicode, line 81: duped from mod_loader

addons/mod_loader_tools/mod_loader_helper.gd
```diff
@@ -0,0 +1,123 @@
...
+	var mod_enabled_state_dict: Dictionary = get_enabled_state_dict()
+	mod_enabled_state_dict[mod_id] = not is_mod_enabled(mod_id)
+	ProjectSettings.set_setting(settings.MOD_ENABLED_STATE_DICT, mod_enabled_state_dict)
+	ProjectSettings.save_custom(get_override_path())
+
+
+# 3.x does not support ligatures (eg. "tag" being replaced by the glyph)
+# so the glyph is used directly (but it's invisible in the code editor :/ )
+# the name is the same as used in font awesome
+# refer to https://fontawesome.com/search?o=r&m=free&s=solid&f=classic
```
ithinkandicode, line 111: Having things that are invisible in the editor isn't really desirable as it can easily cause issues

addons/mod_loader_tools/mod_loader_setup_helper.gd
```diff
@@ -0,0 +1,139 @@
...
+			ProjectSettings.get_setting("display/window/size/height")
+		)
+	)
+	change_scene(ProjectSettings.get_setting(modloaderhelper.settings.GAME_MAIN_SCENE))
+
+
+func try_setup_modloader() -> void:
+	# no more setup needed if the game is already modded
+	if modloaderhelper.are_mods_enabled():
+		print("Mods are enabled.")
```
ithinkandicode, line 91: I think the practice we establish with mod_loader's logs is a good one, instead of using `print` for free text

addons/mod_loader_tools/mod_loader_setup_helper.gd
```diff
@@ -0,0 +1,139 @@
...
+			classes.append(new_class)
+			class_icons[new_class.class] = "" # empty icon, does not matter
+
+	ProjectSettings.set_setting(modloaderhelper.settings.GLOBAL_SCRIPT_CLASSES, classes)
+	ProjectSettings.set_setting(modloaderhelper.settings.GLOBAL_SCRIPT_CLASS_ICONS, class_icons)
+
+	# rename application to make it clear to the user that they are playing modded
+	# godot also creates a new user:// folder for logs (and other things using user:// for saves etc.)
+	var original_name: String = ProjectSettings.get_setting(modloaderhelper.settings.APPLICATION_NAME)
+	ProjectSettings.set_setting(modloaderhelper.settings.APPLICATION_NAME, original_name + " (Modded)")
```
ithinkandicode, line 128: Definitely not a fan of this, users will think their progress has been wiped (as they do for Brotato mods like Assassin, which disables Steam). It also disables Steam saves, which we shouldn't do. Using mods shouldn't affect how the game loads your save file.

It's a good idea in theory, but we've seen how this plays out in practice with Brotato.

addons/mod_loader_tools/mod_loader_setup_helper.gd
```diff
@@ -0,0 +1,139 @@
...
+
+func try_setup_modloader() -> void:
+	# no more setup needed if the game is already modded
+	if modloaderhelper.are_mods_enabled():
+		print("Mods are enabled.")
+		return
+
+	# game needs to be restarted before mods can be loaded
+	# if initialized and not mods enabled yet,
+	# prompt the user to quit and restart in the selector scene
```
ithinkandicode, line 96: The comments in general might benefit from more grammar, incl. capitalisation and periods. In this case, I'm not sure if the comment reads like this:

> Game needs to be restarted before mods can be loaded if initialized and not mods enabled yet.
> Prompt the user to quit and restart in the selector scene

Or this:

> Game needs to be restarted before mods can be loaded.
> If initialized and not mods enabled yet, prompt the user to quit and restart in the selector scene

It doesn't have too much effect on this particular comment, but generally the comments throughout would benefit from this clarity

ithinkandicode (commented):
The code itself looks fine. Most of my notes are for the README. There's also notes on duped code and making comments clearer. The only major issue is using a separate dir for the save files, which would disable Steam saves and has shown to be confusing for users.

ithinkandicode:
Notes on the GUI itself (see screenshots in this comment above)

README.md
```diff
@@ -2,9 +2,137 @@
...
+<!-- TOC -->
+
+
+## For Mod Users
+### Mod Loader Setup
+If the game you want to mod does not natively use this ModLoader, you will have to complete two steps to set it up:
+1. Place the `/addons` folder from the ModLoader next to the executable of the game you want to mod.
+2. set this flag `--script addons/mod_loader_tools/mod_loader_setup_helper.gd`
+   - Steam: right-click the game in the game list > press `properties` > enter in `startup options` 
+   - Godot: Project Settings > press `Editor` > enter in `Main Run Args`
```
ithinkandicode, line 36: Not sure about using backticks so liberally. They've been used in this edit for code, filenames, and general text. Better to have a more flexible system. Eg my approach when writing markdown is:

- *italics* for filenames -- eg "Open *foobar.txt*"
- *italics* for GUI labels -- eg "Click *File* > *Save As* > Enter your filename"
- **bold** for important text -- eg "Saving may **overwrite** progress"
- `backticks` for code and GUI inputs - eg "Save as `foobar.txt`"

Qubus0:
@ithinkandicode we should merge your pr first, then adjust and merge this one on top. 

let's keep the whole readme out of this and reconsolidate it into a proper wiki with separate pages for everything. I have created an issue for that where we can discuss everything: #35. I'll resolve all the comments about the readme here so we can focus on the code
What we do with the Interface branch is also a thing we can decide after that

ithinkandicode:
@Qubus0 I've copied over the content from the README in this PR to a temporary wiki page:

https://github.com/GodotModding/godot-mod-loader/wiki/@TEMP:-PR33

This ensures we keep your new content safe, which should help you fix any merge conflicts with the newer README from the main branch.

addons/mod_loader_tools/mod_loader_helper.gd
ithinkandicode: Is *mod_loader_helper.gd* called before *mod_loader.gd*? I can understand why these methods would be duped if that's the case.

Would it be viable to make another class, eg "*mod_loader_utils*" or something, and have that as the top-most autoload? It could contain the duped methods for stuff like getting CLI args, probably logging too, which would have the extra benefit of making the methods in *mod_loader* focused entirely on just loading mods.

Qubus0: ModLoader is not available in the setup helper before it is, well, set up. That also means having another autoload won't fix it. I even have to use the classes directly from the script because the class isn't registered as global yet.
```gdscript
# mod_loader_setup_helper.gd

# IMPORTANT: use the ModLoaderHelper through this in this script! Otherwise
# script compilation will break on first load since the class is not defined
var modloaderhelper: Node = load("res://addons/mod_loader_tools/mod_loader_helper.gd").new()

var skip_mod_selection: bool = modloaderhelper.should_skip_mod_selection()

# mod_loader_helper.gd
static func should_skip_mod_selection() -> bool:
	return cmd_line_arg_exists("--skip-mod-selection") or\
		is_project_setting_true(settings.SKIP_MOD_SELECTION)
```

ithinkandicode: Ah, I see what you mean. *mod_loader_helper.gd* is being run via the the `--script` arg so nothing can be autoloaded before it. This convo is resolved for me then 👍

addons/mod_loader_tools/mod_loader.gd
```diff
@@ -112,7 +113,7 @@ func _init():
 	# Loop over "res://mods" and add any mod zips to the unpacked virtual
 	# directory (UNPACKED_DIR)
 	_load_mod_zips()
-	mod_log("DONE: Unziped all Mods", LOG_NAME)
+	mod_log("DONE: Loaded all mod files", LOG_NAME)
```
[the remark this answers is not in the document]
ithinkandicode: I'm just gonna resolve this as my comment is a bit of a nitpick, no need to hold up the approval due to this 👍

addons/mod_loader_tools/mod_loader_helper.gd
```diff
@@ -0,0 +1,123 @@
...
+
+# Required keys in a mod's manifest.json file
+const REQUIRED_MANIFEST_KEYS_ROOT = [
+	"name",
+	"version_number",
+	"website_url",
+	"description",
+	"dependencies",
+	"extra",
+]
```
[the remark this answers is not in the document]
ithinkandicode: +1 for a new class to handle shared code, I'll mark this as resolved

addons/mod_loader_tools/mod_loader_helper.gd
ithinkandicode: So if I understand it correctly, mod_loader_helper is a set of static utilities, is that right?

It seems like this should be a larger, separate PR: One that makes such a utility class, and moves anything that's potentially duplicatable out of *mod_loader* and into its utility class

I think the filenames could be shortened too. Eg *"mod_loader_setup_helper"* could just be called "autosetup.gd", and *"mod_loader_helper"* could just be *"utils.gd"*.

addons/mod_loader_tools/mod_loader_helper.gd
ithinkandicode: Contributors may be editing these files in Godot, rather than an external editor, so I think making things invisible could be a point of impasse here. Is there an alternative implementation to whatever FontAwesome is used for here?

ithinkandicode:
So I think that what's being done in this PR would benefit from being split into smaller PRs. Atm you have:

1. The auto setup stuff (*setup_helper*)
2. A utility file, with some duped stuff from *mod_loader.gd* (*helper*)
3. A mod class, with its own validation (*mod_details*)
4. Code that adds support for the GUI in [#36](https://github.com/GodotModding/godot-mod-loader/pull/36)

I think you could split this into four PRs, one for each of the above

ithinkandicode:
RE: *override.cfg* is used: What happens when the game updates? It looks like the *override.cfg* that got created when I ran the setup for Brotato includes all the vanilla stuff. Can you use an *override.cfg* that only includes certain settings, or does it have to be the entire project?

If it has to be the entire project's configs, perhaps could you check to see if the game has been updated since *override.cfg* was created? Eg. you could create a backup of the vanilla *override.cfg* during the initial setup phase, then you could re-regenerate it from vanilla again later and check its checksum. That'd show you if anything's been changed in vanilla

*(note: I think I said this elsewhere but can't find it)*

Qubus0:
oh yep good idea. the interface only covers manual resets so far

Qubus0:
reworking this into multiple new prs

ithinkandicode:
> oh yep good idea. the interface only covers manual resets so far

Cool. I mentioned saving a copy of the vanilla *override.cfg*, just noting here that it might be best to save it in `user://` so it doesn't bloat the game dir too much.

PR #36, opened 2023-01-16
Only Interface PR

Qubus0:
split the pr #33 into two parts for only self installation and only UI.
the loader now falls back to default behaviour if the `interface` folder is not present. that allows the user to just deselect that folder when installing from the asset lib

to test:
1. Redoing the loader setup via button and restarting
2. Toggling mods on and off (use these two small ones that only print stuff for you to test with: [download mods.zip](https://github.com/GodotModding/godot-mod-loader/files/10421417/mods.zip)). A disabled mod should not be initialised upon game start.
3. skipping the selection screen (via flag, pressing <kbd>ALT</kbd> or via the option in the selection)

Stuff that's not done yet:
1. mod icons
2. flag for autoload mods that are always loaded right away (state can only be changed when the game is restarted)


ithinkandicode:
*Migrated screenshots from PR #33:*

<details>
<summary><b>Screenshots</b></summary>

Initial Setup - always shown on first load:



After initial setup, with no mods in `./mods`. Starts in fullscreen:



After adding mods and restarting:



After disabling the fullscreen through Brotato's own options, restarting, and enabling 1 mod:


</details>

ithinkandicode:
*Migrated comments from PR #33:*

Notes on the GUI itself (see screenshots in [this comment](#issuecomment-1384637200) above):

- Sizes:
  - Text is way too small. Some text is unreadable
  - Mod icons are too big compared to their text. At the non-fullscreen size you can only see 2 mods at once
    - Maybe add a toggle to change the icon size? Eg layout buttons for List - Large/List - Compact/Grid  
- Fullscreen:
  - Starts in fullscreen, so needs an option to toggle fullscreen after starting
  - If it starts without fullscreen (eg screenshot 4), some of the GUI is cut off, and the author string is malformed (maybe just move it to a separate line?)
- override.cfg:
  - I don't fully understand how using *override.cfg* works, but won't it cause the game to break if it gets updated? 
  - If so, there would need to be some way to check that *override.cfg* hasn't become outdated 
  - Eg. diff the custom one vs. the vanilla one -- maybe you could save a backup of the vanilla one and check its checksum?
- Needs dependency checks, and to show dependencies

otDan:
I still think this should be a separate repo fully separated from the mod loader

Qubus0:
then there will be a few bits of code dotted around the setup helper that don't make any sense since the interface does not exist, but which are required to be in those places for it to work. sadly the coupling itself is not very dynamically extendable there

otDan:
The problem is that when you want to update stuff you will have to keep maintaining everything and not just the interface part, the more modularity the better workflow we can have

Qubus0:
sure, I doubt there will be much to maintain in the interface since it's just a way to display the data parsed into ModDetails

Qubus0:
closing this until self installation is done

PR #49, opened 2023-01-16
improve date padding and formatting for logs

Qubus0:


ithinkandicode (approved):
Approved!

loader/mod_loader.gd
```diff
@@ -186,30 +186,27 @@ func mod_log(text:String, mod_name:String = "Unknown-Mod", pretty:bool = false)-
...
 	if(!log_file.file_exists(MOD_LOG_PATH)):
 		log_file.open(MOD_LOG_PATH, File.WRITE)
-		log_file.store_string("\n" + str(date_time_string,'   ', 'Created mod.log!'))
+		log_file.store_string('%s    Created mod.log!' % date_time_string)
 		log_file.close()
 
 	var _error = log_file.open(MOD_LOG_PATH, File.READ_WRITE)
-	if(_error):
+	if _error:
 		print(_error)
```
ithinkandicode, line 206: I noticed you used `printerr` in another PR, would it be suitable to use it here?

Qubus0: probably better to just use mod log with error levels later anyway

PR #50, opened 2023-01-16
move mod_loader into an addon folder

Qubus0:


ithinkandicode (approved):
Approved! 👍

otDan (commented):
Is there a reason to call the folder mod_loader_tools instead of just mod_loader?

Qubus0:
@otDan 
> with the other things in there and a modularity with the interface for example it made more sense to call it a toolset, for me at least
and I feel like we'll be expanding on this thing as well

ithinkandicode:
I do think *mod_loader* is cleaner. "Mod Loader Tools" suggests that it's a set of tools for the mod loader, rather than the mod loader itself

otDan:
> I do think _mod_loader_ is cleaner. "Mod Loader Tools" suggests that it's a set of tools for the mod loader, rather than the mod loader itself

I agree with this

KANAjetzt:
What do we do with ``run.gd``? 
Its a bit out of place.

Should be add a *"game_dir_copy"* folder? 
Until we can get rid of ``run.gd``.

ithinkandicode (approved):
Approved

KANAjetzt (approved):
lets go 🔥

Qubus0:
run.gd won't be around for long. new flag is `--script addons/mod_loader/run.gd`

PR #60, opened 2023-01-17
serialize into ModData and ModDetails 

### LINKED ISSUE #54: validate mod ids completely
as per the info by thunderstore, a "full id" would be `{namespace}-{name}-{version_number}`
I'm not sure what we would call this id, maybe something like "full_name_id"?

the validation does address the composition, but not the elements of that composition. following thunderstore,
namespace and name are are validated with this `^[a-zA-Z0-9_]$` (where _ can only be replaced with space for display purposes)
and the version_number with this `^[0-9]+\.[0-9]+\.[0-9]+$`
Godot does have a regex class you can use

_Originally posted by @Qubus0 in https://github.com/GodotModding/godot-mod-loader/pull/51#pullrequestreview-1250623458_

Qubus0:
for better type safety and convenience

this is only internal, nothing needs to be changed in mods.

@ithinkandicode check if this works for you to also integrate the config stuff

also closes #54

### DIFF
--- a/addons/mod_loader/mod_details.gd
+++ b/addons/mod_loader/mod_details.gd
@@ -1,0 +1,132 @@
+extends Resource
+class_name ModDetails
+
+var name := ""
+var namespace := ""
+var version_number := "v0.0.0"
+var description := ""
+var website_url := ""
+var dependencies := []				# Array[String]
+
+var authors := [] 					# Array[String]
+var compatible_game_version := [] 	# Array[String]
+var tags := [] 						# Array[String]
+var description_rich := ""
+var incompatibilities := [] 			# Array[String]
+var image: StreamTexture
+
+## Required keys in a mod's manifest.json file
+const REQUIRED_MANIFEST_KEYS_ROOT = [
+	"name",
+	"namespace",
+	"version_number",
+	"website_url",
+	"description",
+	"dependencies",
+	"extra",
+]
+
+## Required keys in manifest's `json.extra.godot`
+const REQUIRED_MANIFEST_KEYS_EXTRA = [
+	"authors",
+	"compatible_mod_loader_version",
+	"compatible_game_version",
+	"incompatibilities",
+	"config_defaults",
+]
+
+
+func _init(manifest: Dictionary) -> void:
+	if (not dict_has_fields(manifest, REQUIRED_MANIFEST_KEYS_ROOT) or
+		not dict_has_fields(manifest.extra, ["godot"]) or
+		not dict_has_fields(manifest.extra.godot, REQUIRED_MANIFEST_KEYS_EXTRA)):
+			return
+
+	name = manifest.name
+	namespace = manifest.namespace
+	version_number = manifest.version_number
+	if (not is_name_or_namespace_valid(name) or
+		not is_name_or_namespace_valid(namespace) or
+		not is_semver_valid(version_number)):
+			return
+
+
+	description = manifest.description
+	website_url = manifest.website_url
+	dependencies = manifest.dependencies
+
+	var 	godot_details: Dictionary = manifest.extra.godot
+	authors = _get_array_from_dict(godot_details, "authors")
+	incompatibilities = _get_array_from_dict(godot_details, "incompatibilities")
+	compatible_game_version = _get_array_from_dict(godot_details, "compatible_game_version")
+	description_rich = _get_string_from_dict(godot_details, "description_rich")
+	tags = _get_array_from_dict(godot_details, "tags")
+
+	# todo load file named icon.png when loading mods and use here
+#	image StreamTexture
+
+
+func get_mod_id() -> String:
+	return "%s-%s" % [namespace, name]
+
+
+func get_package_id() -> String:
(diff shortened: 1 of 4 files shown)


ithinkandicode:
What's the difference between *mod_details.gd* and *mod_data.gd*? It looks like there's code for validation in both, would all the mod-specific stuff be better in a single file instead of split across two?

It might also be helpful to include a comment at the top of those files stating what their purpose is. Generally I think all funcs can benefit from comments, esp. if you're not familiar with the codebase.

For static funcs, would they be better being visually separated from the non-static funcs? Ie. you could move them down the file into their own section with comment headings. Eg. I used this approach throughout *mod_loader.gd* (eg [here](https://github.com/GodotModding/godot-mod-loader/blob/main/addons/mod_loader/mod_loader.gd#L76)):

```
# Static Funcs
```

Qubus0:
oh totally forgot to sort them again. will also add comments
ModDetails stores and validates all the content from the manifest, while ModData stores and validates everything needed to properly load the mod

addons/mod_loader/mod_details.gd
```diff
@@ -0,0 +1,160 @@
...
+
+## A valid namespace may only use letters (any case), numbers and underscores
+## and has to be longer than 3 characters
+## /^[a-zA-Z0-9_]{3,}$/
+static func is_name_or_namespace_valid(name: String) -> bool:
+	var re := RegEx.new()
+	re.compile("^[a-zA-Z0-9_]*$") # alphanumeric and _
+
+	if re.search(name) == null:
+		printerr('Invalid name or namespace: "%s". You may only use letters, numbers and underscores.' % name)
```
ithinkandicode, line 102: Is it possible to use `mod_log` for logging? I think it would be best if we keep all logging through ModLoader allocated to just the logger funcs, for consistency in both the code and the things added to *godot.log*

addons/mod_loader/mod_details.gd
```diff
@@ -0,0 +1,160 @@
...
+## A valid semantic version should follow this format: {mayor}.{minor}.{patch}
+## reference https://semver.org/ for details
+## /^[0-9]+\\.[0-9]+\\.[0-9]+$/
+static func is_semver_valid(version_number: String) -> bool:
+	var re := RegEx.new()
+	re.compile("^[0-9]+\\.[0-9]+\\.[0-9]+$")
+
+	if re.search(version_number) == null:
+		printerr('Invalid semantic version: "%s". ' +
+		'You may only use numbers and periods in this format {mayor}.{minor}.{patch}' % version_number)
```
ithinkandicode, line 122: Is it possible to show the mod name? That would make debugging much easier, esp. for logs like this which atm don't tell you which mod they're for

addons/mod_loader/mod_details.gd
```diff
@@ -0,0 +1,160 @@
+extends Resource
+## Stores and validates contents of the manifest set by the user
+class_name ModDetails
```
ithinkandicode, line 3: Perhaps this file/class would be better named as "mod_manifest"/`ModManifest`?

ithinkandicode (commented):
**This is amazing work, well done Ste!** 🏆

I only have a few minor points: Potentially change a class name, and make the logging consistent

addons/mod_loader/mod_details.gd
Qubus0: yep, will do that. I think i'll merge the logger/util improvements on top of this one

addons/mod_loader/mod_details.gd
Qubus0: good point, will do

addons/mod_loader/mod_details.gd
Qubus0: done

Qubus0:
will wait with merging until we have a logger class with static methods to log from, as those seem to work in engine and exported and with automatic setup.

ithinkandicode (approved):
Looks good to me. Like you said, there's a few bits left to do, but they can be raised in a subsequent PR

ithinkandicode:
@Qubus0 I don't think we should use godot 4 [documentation comments](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_documentation_comments.html) yet. The VS Code plugin for Godot already shows comments, but it doesn't handle the `##` version correctly.

Eg this uses the normal `#` comment:

docstring-1

But it becomes this if you use `##` for comments:

docstring-2

So I think we should hold off on using them until the plugin is fixed

Qubus0:
hm that's a bummer. I guess it also doesn't handle the bbcode references to classes and methods? Looks like it's treating two ## as markdown heading

ithinkandicode (approved):
I think everything I mentioned before has been sorted here

And again, really great work Ste :)

MythicManiac:
Does this account for leading zeroes? e.g. `01.01.01` should probably be disallowed

Qubus0:
@MythicManiac not yet, but I can add it. Probably should also disallow numbers over 999, since right now versioning a la LaTeX is still possible (0.3.1415, adding a digit of pi for every update)

PR #62, opened 2023-01-19
Split off logging and utils

### LINKED ISSUE #21: Different log levels for mod_log
> Ste — Today at 16:23
> should we have different logs levels for mod_log? INFO, WARNING, ERROR and just add those at the front of the line? would make reading the logs easier if you use a highlighter

Something like:

`11.1.2023 - 08:28:20   ModLoader-INFO: Initializing -> Cube-HarvestCalc`

Qubus0:
This needs to be merged after 
- #60 

Breaking changes:
Removed:
```
ModLoader.
mod_log
dev_log
```

Use the following instead:
```
ModLoaderUtils.
log_fatal()
log_error()
log_warning()
log_info()
log_success()
log_debug()
log_debug_json_print()
```
closes #21

### DIFF
--- a/addons/mod_loader/mod_data.gd
+++ b/addons/mod_loader/mod_data.gd
@@ -40,13 +40,13 @@ func load_manifest() -> void:
 	if not has_required_files():
 		return
 
-#	ModLoader.mod_log("Loading mod_manifest (manifest.json) for -> %s" % dir_name, LOG_NAME)
+	ModLoaderUtils.log_info("Loading mod_manifest (manifest.json) for -> %s" % dir_name, LOG_NAME)
 
 	# Load meta data file
 	var manifest_path = get_required_mod_file_path(required_mod_files.MANIFEST)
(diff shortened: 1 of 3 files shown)


otDan:
probably switching back to the single method and just adding an enum to specify the log type is better

Qubus0:
Any particular reason for that? That would mean having to reference the enum every single time for logging, e.g.
```gdscript
ModLoaderUtils.mod_log("some message here: %s" % some_value, ModLoaderUtils.verbosity_level.INFO, LOG_NAME)
# and this if you also want the json print included in the same method
ModLoaderUtils.mod_log("some message here: %s" % some_value, ModLoaderUtils.verbosity_level.INFO, LOG_NAME, true)
```
compared to this.
```gdscript
ModLoaderUtils.log_info("some message here: %s" % some_value, LOG_NAME)
```
I feel like it's much more concise and better to understand at a glance

otDan (approved):
well other than that comment, everything seems to be in place and working

ithinkandicode (approved):
Lookin' good man!

PR #63, opened 2023-01-19
Move other utils

Qubus0:
Needs to be merged after 
- #62 

No breaking changes

Moves other utility functions from ModLoader and ModData to new ModLoaderUtils class. 
Only affects internal methods


ithinkandicode (approved):
👌

PR #69, opened 2023-01-20
Improve static typing and style

Qubus0:
Lots of little things to make it easier to avoid issues in the future.
Please test this with your mods, it should not have any effect at all.
Let me know if I missed something

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -29,57 +29,57 @@ extends Node
 # If true, a complete array of filepaths is stored for each mod. This is
 # disabled by default because the operation can be very expensive, but may
 # be useful for debugging
-const DEBUG_ENABLE_STORING_FILEPATHS = false
+const DEBUG_ENABLE_STORING_FILEPATHS := false
 
 # Path to the mod log file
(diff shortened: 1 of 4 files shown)


addons/mod_loader/mod_loader.gd
```diff
@@ -497,37 +493,37 @@ func save_scene(modified_scene, scene_path:String):
...
+	var error_msg := ""
+	var data = {} # can be anything
+	var defaults := {}
 
 	# Invalid mod ID
-	if !mod_data.has(mod_id):
+	if not mod_data.has(mod_dir_name):
 		error_num = 1
-		error_msg = str("ERROR - Mod ID was invalid: ", mod_id)
+		error_msg = "ERROR - Mod ID was invalid: " % mod_dir_name
```
otDan, line 505: missing from string

Qubus0: good catch! fixed

KANAjetzt:
I think #70 is missing on this one - right?

Qubus0:
@KANAjetzt develop is missing #70 as well, since it got merged past

ithinkandicode (approved):
👍

PR #70, opened 2023-01-20
fix log-ignore flag

Qubus0:
should cover no flag and any flag. spaces still do not work due to godot splitting them.
new format: `--log-ignore=ModLoader,ModLoader:ModData,TestMod` (no spaces, no quotation marks)


KANAjetzt (approved):
🔨

PR #75, opened 2023-01-23
Self setup

### LINKED ISSUE #8: Mod Loader without editing source code
> **Ste - *January 3, 2023 11:37 AM***
Hey btw, back to the idea of only needing the modloader and not some edits to the source code (like adding it as auto load): godot can override project settings with a override.cfg placed in the same directory as the exe. One problem being that it completely replaces it and can’t just ‘add to it’.. though there might be a way, since you can save the project settings as a custom file ([https://discord.com/channels/212250894228652034/342056330523049988/873285063805124678](https://discord.com/channels/212250894228652034/342056330523049988/873285063805124678))
> 
> 
> > I can save these config settings very easily using the following:
> > 
> > 
> > ```
> > var path = ProjectSettings.globalize_path(OS.get_executable_path().get_basename() + "override.cfg")
> > ProjectSettings.save_custom(path)
> > ```
> > 
> 
> That means we could possibly run the project once, make it add itself to the auto loads, create the override and close again
> 

> **Ste - *January 3, 2023 11:51 AM***
Adding it as auto load might be a bit annoying though, since it is added as last in the list. So get existing, hold them, remove them, add loader, add them back..
also, save_custom saves all of the settings, which might be bad.. and projects using this mechanic might be affected (though i've never seen a project use that)
>

Qubus0:
Use this startup flag to enable the self setup `--script addons/mod_loader/mod_loader_setup.gd`
Changes will apply after one restart. Setup was successful when the window name shows `(Modded)`
If the ModLoader autoload is already registered in the ProjectSettings, setup will be skipped - remove it when testing the setup.

As discussed before and in this [Godot proposal #6137](https://github.com/godotengine/godot-proposals/discussions/6137) here, the mod loader will end up as the last autoload when set up this way. 
That is completely fine for all mods that do not modify autoloads.
Mods that do modify autoloads will not work properly. To use those mods, you will still have to decompile the game and add it as the first autoload manually.

salvaged from #33
closes #8

### DIFF
--- a/addons/mod_loader/mod_loader_setup.gd
+++ b/addons/mod_loader/mod_loader_setup.gd
@@ -78,16 +78,21 @@ func setup_modloader() -> void:
 	modloaderutils.log_info("ModLoader setup complete", LOG_NAME)
 
 
+func is_loader_set_up() -> bool:
+	return is_project_setting_true(settings.IS_LOADER_SET_UP)
+
+
+func is_loader_setup_applied() -> bool:
+	if not root.get_node_or_null("/root/ModLoader") == null:
+		if not is_project_setting_true(settings.IS_LOADER_SETUP_APPLIED):
+			modloaderutils.log_info("ModLoader is already set up. No self setup required.", LOG_NAME)
+		return true
+	return false
+
+
 static func is_project_setting_true(project_setting: String) -> bool:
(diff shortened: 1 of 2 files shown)


ithinkandicode (approved):
All looks good to me, I think I looked over this before and saw no issues

Fantastic work as always Ste :)

PR #80, opened 2023-01-24
Overwrite the autoload order in the ProjectSettings

otDan:
We need this so the mod loader can setup itself as first in the autoload list, so we do not distribute PCKs with game files to violate game copyrights


otDan:
Closing in favour of the new binary replace

PR #81, opened 2023-01-27
Project binary injection

### LINKED ISSUE #34: automatically create mods and mods-unpacked directories
for ease of setup for everyone

KANAjetzt:
# Project binary injection

Because we cant set the order of autoloads via ``override.cfg`` we have to use external tooling ( [GodotPckTool](https://github.com/hhyyrylainen/GodotPckTool) ) to inject the modified project settings in to the games .pck file.

*Relevant feature proposal:*
https://github.com/godotengine/godot-proposals/discussions/6137

## OS.execute() - GodotPckTool
```python
# Add modified binary to the pck
func inject_project_binary() -> void:
	var output_add_project_binary := []
	var _exit_code_add_project_binary := OS.execute(path.pck_tool, ["--pack", path.pck, "--action", "add", "--file", path.project_binary, "--remove-prefix", path.mod_loader_dir], true, output_add_project_binary)
	modloaderutils.log_debug_json_print("Adding custom project.binaray to res://", output_add_project_binary, LOG_NAME)
```

This is the main part of this PR.   
Executing the *GodotPckTool*, adding the before created ``project.binary`` to the main ``.pck``.
```python
	# save the current project settings to a new project.binary
	ProjectSettings.save_custom(exe_path + "addons/mod_loader/project.binary")
```

### --silent
Executes the setup with no user interaction.
*Restarts the game immediately after the setup, without showing the countdown*

### --setup-create-override-cfg
Forces the setup to create the ``override.cfg`` in the game base directory.
Skips the ``project.binary`` injection.
*The setup checks if there is a ``exe_name.pck`` file in the game base directory and decides based on this if the injection can happen. So this argument is only required if this behavior is not wanted.*

### --exe-name="name"
Override the ``exe_name``.

### --pck-name="name"
Override the ``pck_name``.

---

## Things this PR not covers

- Installation on OSX
- Installation on Linux
- Installation for embedded ``.pck`` files

---
closes #34


addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
...
 	if is_loader_set_up() and not is_loader_setup_applied():
-		modloaderutils.log_info("ModLoader is set up, but the game needs to be restarted", LOG_NAME)
-		OS.alert("The Godot ModLoader has been set up. Restart the game to apply the changes. Confirm to quit.")
+		modloaderutils.log_info("ModLoader is set up, the game will be restarted", LOG_NAME)
+
 		ProjectSettings.set_setting(settings.IS_LOADER_SETUP_APPLIED, true)
 		ProjectSettings.save_custom(modloaderutils.get_override_path())
+
+		# run the game again to apply the changed project settings
+		var _exit_code_game_start = OS.execute(exe_path + "Brotato.exe", ["--script", "addons/mod_loader/mod_loader_setup.gd", '--pck-name="Brotato.pck"', "--log-debug"], false)
```
Qubus0, line 116: good start, but this won't work on any other OS or on any other game

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
 		OS.set_window_title("%s (Modded)" % ProjectSettings.get_setting("application/config/name"))
 		return
 
+	var pck_name : String = modloaderutils.get_cmd_line_arg_value("--pck-name")
+	var exe_path : String = modloaderutils.get_local_folder_dir()
+	modloaderutils.log_debug("exe_path -> " + exe_path, LOG_NAME)
+
+# remove and re-add autoloads
```
ithinkandicode, line 54: missing indent

addons/mod_loader/godotpcktool/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
ithinkandicode, line 1: Maybe *godotpcktool* should be in a `vendor` subdirectory? Esp. if we might need to add [GodotPCKExplorer](https://github.com/DmitriySalnikov/GodotPCKExplorer) later for splitting PCKs out of EXEs

addons/mod_loader/mod_loader_setup.gd
ithinkandicode, line 54: missing indent before this comment

addons/mod_loader/mod_loader.gd
```diff
@@ -89,6 +89,16 @@ func _init() -> void:
...
 
+	# Log Autoload order
+	var autoloads := {}
+	for prop in ProjectSettings.get_property_list():
+			var name: String = prop.name
+			if name.begins_with("autoload/"):
+					var value: String = ProjectSettings.get_setting(name)
+					autoloads[name] = value
+
+	ModLoaderUtils.log_debug_json_print("Autoload order", autoloads, LOG_NAME)
```
ithinkandicode, line 100: Do we need to log the autoloads? I agree that it's generally useful info but I'm not sure if it's needed here in ModLoader. 

If it's only here so we can check that ModLoader is the first autoload, perhaps it would be more suitable to just log when it's not first, as that's the only bit of info we're concerned with?

And in those cases, we may want to restart the setup process anyway?

- Note: If we do take the approach of "if not first autoload, run setup" then we will need to be careful, because if a bug occurs that technically completes the setup but doesn't successfully add ModLoader as the first autoload, then we could enter an infinite loop, ie:
- *setup > bug occurs but setup completes > our check sees that modloader isn't the first autoload >*
*setup again > bug occurs again > etc*

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -34,6 +34,7 @@ var modloaderutils: Node = load("res://addons/mod_loader/mod_loader_utils.gd").n
 
 
 func _init() -> void:
+	modloaderutils.log_debug("Mod-Loader setup initialized", LOG_NAME)
```
ithinkandicode, line 37: I think we're using `ModLoader` rather than `Mod-Loader` in the comments that refer to it

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
 		OS.set_window_title("%s (Modded)" % ProjectSettings.get_setting("application/config/name"))
 		return
 
+	var pck_name : String = modloaderutils.get_cmd_line_arg_value("--pck-name")
```
ithinkandicode, line 50: Is the PCK name not always the EXE name? And isn't the EXE name always `application/config/name`?

If *no* to either of those, is there a better way to target the PCK? I'm asking because if you need to target the PCK by name, then the setup instructions for every game will be different. Whereas, if you don't need to target the name, then the CLI arg is universal.

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
...
+	modloaderutils.log_debug("pck_tool_path -> " + pck_tool_path, LOG_NAME)
+
+	# TODO: Pck potentially embedded in the games .exe
+	# C:/path/to/game/game.pck
+	var pck_path := exe_path + pck_name
+	modloaderutils.log_debug("pck_path -> " + pck_path, LOG_NAME)
+
+	# C:/path/to/game/addons/mod_loader/project.binary
+	var project_binary_path := mod_loader_dir_path + "/project.binary"
+	modloaderutils.log_debug("project_binary_path -> " + project_binary_path, LOG_NAME)
```
ithinkandicode, line 90: What do you think about aligning these, to make it easier to read them in the logs. And as they're logged in a group, perhaps the logging could be done all at once too, to better separate variable declaration from logging? Eg:

```gdscript
# Log all paths
modloaderutils.log_debug("mod_loader_dir_path -> " + mod_loader_dir_path, LOG_NAME)
modloaderutils.log_debug("pck_tool_path       -> " + pck_tool_path, LOG_NAME)
modloaderutils.log_debug("pck_path            -> " + pck_path, LOG_NAME)
modloaderutils.log_debug("project_binary_path -> " + project_binary_path, LOG_NAME)
```

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
...
+
+	# Create a backup of the original pck
+	var output_backup_pck := []
+	var _exit_code_backup_pck := OS.execute(pck_tool_path, ["--pack", pck_path, "--action", "repack", " " + pck_name + "_backup"])
+	modloaderutils.log_debug(output_backup_pck, LOG_NAME)
+	modloaderutils.log_debug_json_print("Creating a backup of the original pck", output_backup_pck, LOG_NAME)
```
ithinkandicode, line 96: We do we need to make a backup of the vanilla PCK? 

I'm not sure what problem this solves, and it can be a time consuming operation to perform. So unless having a backup of vanilla PCK is mandatory to the process (and I don't think it is), then we can skip doing that, which will make the initial setup go much faster.

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
...
+
 	setup_modloader()
 
 	# If the loader is set up, but the override is not applied yet,
 	# prompt the user to quit and restart the game.
 	if is_loader_set_up() and not is_loader_setup_applied():
-		modloaderutils.log_info("ModLoader is set up, but the game needs to be restarted", LOG_NAME)
-		OS.alert("The Godot ModLoader has been set up. Restart the game to apply the changes. Confirm to quit.")
+		modloaderutils.log_info("ModLoader is set up, the game will be restarted", LOG_NAME)
+
```
ithinkandicode, line 111: We are we hiding the restart from users? Please consider this from their perspective: They started the game and it suddenly stopped. From the user's perspective, that's a crash. Ste's approach is much cleaner IMO. 

And I can see from the code below this (which Ste has commented on) that the workaround required to do this currently requires hardcoding the game's EXE name into ModLoader (which as Ste pointed out, is not viable).

If you don't want to notify users, then I would advise simply adding a CLI arg for that, eg. `--silent` which is a commonly used CLI arg to do this.

addons/mod_loader/mod_loader_setup.gd
ithinkandicode, line 116: @Qubus0 is correct, we can't hardcode the EXE name. 

Regarding the starting of the game only working on Windows, I think that's fine for the MVP if this only happens when using the `--silent` CLI arg (see my other review comment [here](https://github.com/GodotModding/godot-mod-loader/pull/81#discussion_r1089406758)), because we can just say "silent installation is currently only supported on Windows" -- then we can expand that later.  This would let us, for example, allow silent installations on Thunderstore by default, if we choose to do that.

As long as it's an *opt-in* to run platform-specific code, we won't be preventing users who aren't on that platform from installing/running ModLoader, and that's fine for the MVP. Just be sure to keep the docs updated, perhaps with a dedicated "Platform Specifics" page to maximise its visability ;)

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -46,15 +47,75 @@ func try_setup_modloader() -> void:
...
+	var _exit_code_backup_pck := OS.execute(pck_tool_path, ["--pack", pck_path, "--action", "repack", " " + pck_name + "_backup"])
+	modloaderutils.log_debug(output_backup_pck, LOG_NAME)
+	modloaderutils.log_debug_json_print("Creating a backup of the original pck", output_backup_pck, LOG_NAME)
+
+	# Add modified binary to the pck
+	var output_add_project_binary := []
+	var _exit_code_add_project_binary := OS.execute(pck_tool_path, ["--pack", pck_path, "--action", "add", "--file", project_binary_path, "--remove-prefix", mod_loader_dir_path], true, output_add_project_binary)
+	modloaderutils.log_debug_json_print("Adding custom project.binaray to res://", output_add_project_binary, LOG_NAME)
+
+	# TODO: Remove unnecessary files after installation?
```
ithinkandicode, line 103: Can all this project.binary patch stuff be moved to a dedicated func? That would help contain all the code thats' related to it, making it easier to maintain and read.

We've also been talking on Discord about offering a CLI arg that lets you only save the *override.cfg* file, so separating all this patcher code would let us apply such a CLI arg option much more easily.

ithinkandicode:
@KANAjetzt  We've been talking on Discord about offering a **CLI arg that lets you only save the *override.cfg* file**. Could you include support for this in this PR please? The current use case is Dome Keeper, which has an embedded PCK, so we can't patch it with the methods currently applied in this PR. Eg. the CLI arg could be `--setup-override-cfg-only`

Dome Keeper users *could* use [GodotPCKExplorer](https://github.com/DmitriySalnikov/GodotPCKExplorer) to extract the PCK, but that'll need to be done after every game update, and can't be handled automatically, so this isn't really viable if the user is using Thunderstore. Granted, *override.cfg* has the major limitation of not being able to put ModLoader at the top of the autoloads list (and Dome Keeper does indeed have many autoloads of its own), but @Qubus0 has suggested that this is a reasonable trade-off considering how impractical the alternative would be (ie. repeatedly using GodotPCKExplorer), and that their knowledge of the current scope of Dome Keeper mods suggests that modifying DK's autoloads may not be necessary.

I also think that this is a good thing to support anyway, and will keep the code that generates *override.cfg* in place for if/when Godot implements [Ste's proposal](https://github.com/godotengine/godot-proposals/discussions/6137) to allow changing the autoloads order.

*Note: I also mentioned this approach briefly in [this review comment](https://github.com/GodotModding/godot-mod-loader/pull/81#discussion_r1089533379)*

addons/mod_loader/mod_loader_setup.gd
ithinkandicode: In hindsight, yeah maybe it's better to target the PCK by name, rather than our code just doing things automatically and assuming things. And it's not really an issue to say that in the docs :)

Lemme know your thoughts there tho. Maybe we *try* to guess the PCK name if it's not specified? Looks like the code here just won't work atm if the CLI arg isn't specified. Might be a good idea to mention that the CLI arg is required in the main PR description too, always good to maximise visibility of this kinda stuff 👍

ithinkandicode:
Once this PR is finished and merged, we can look into extending the features here for #83

addons/mod_loader/mod_loader_setup.gd
otDan: indeed it could seem like a crash, but if you actually run it you can see that the restart is pretty much instantaneous, you don't even see the other window close before the new one starts up

Qubus0: you don't see it close because the new window is in front of the old one. the game runs twice at the same time. not sure if that's good.

addons/mod_loader/mod_loader.gd
ithinkandicode: @KANAjetzt It looks like the latest changes remove the autoload logging, which is great, but I was wondering if you considered what I said about adding a log specifically if ModLoader is not the first autoload?

I think that would be very helpful to have, since if it's not the first autoload, then something is probably wrong and mods may/will fail.

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: fixed by using ``OS.get_executable_path()`` 👍

addons/mod_loader/godotpcktool/LICENSE
KANAjetzt: [added ``vendor`` directory 👍](https://github.com/GodotModding/godot-mod-loader/pull/81/commits/1eb0520c6b264a2b49b70f005a0cfeeaf46eaf40)

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: [fixed 👍](https://github.com/GodotModding/godot-mod-loader/pull/81/commits/ee51c3c394ed950f6f4b083df738e0c3b38712bd)

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: [fixed 👍](https://github.com/GodotModding/godot-mod-loader/pull/81/commits/ee51c3c394ed950f6f4b083df738e0c3b38712bd)

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: The installer will use the name from ``OS.get_executable_path()`` if no ``--exe-name="name"`` or ``--pck-name="name"`` cli arg is passed.

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: I added the ``path`` and ``file_name`` dictionary's and log them now with ``modloaderutils.log_debug_json_print()``

https://github.com/GodotModding/godot-mod-loader/pull/81/commits/7a8c627de50f4f36d6b892efba8f6cac35c77b23

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: True - the user can always go back to vanilla by reinstalling the game.

https://github.com/GodotModding/godot-mod-loader/pull/81/commits/f5c2bc64c733d04a0597d1008fd1da9e9f08c559

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: I added a label to inform the user about the restart.
With ``--silent`` the restart will be instand with no user notification.

https://github.com/GodotModding/godot-mod-loader/pull/81/commits/7a8c627de50f4f36d6b892efba8f6cac35c77b23
https://github.com/GodotModding/godot-mod-loader/pull/81/commits/614691d38c4918b61e296367dbe496be964d16bf

addons/mod_loader/mod_loader_setup.gd
KANAjetzt: I fixed the hard coded exe name ( not sure how I missed that one in the first place 👀 )

About the ``--silent``
Currently I have it like this:

```python
func restart_game() -> void:
	# run the game again to apply the changed project settings
	var _exit_code_game_start = OS.execute(path.exe, ["--script", path.mod_loader_dir + "mod_loader_setup.gd", "--log-debug"], false)
	# quit the current execution
	quit()
```

So the game is always executed before quitting the first instance.
What I read here is that the default should be ``OS.alert()`` and leave the restart to the user - right?

In that case I will add ``--unattended`` to show the UI and auto restart.

So we have:

- nor arg -> user handles the restart after ``OS.alert()`` message
- ``--unattended`` user gets notified by restart timer - setup handles "restart"
- ``--silent`` no restart notification - setup immediately "restarts" the game
*From the latest discord chat with Mythic*
- ``--only-setup`` no restart - setup just quits the current game instance

*"restart" in quotes because it's not really a restart, we have to execute the second game instance before we quit the first one.*

addons/mod_loader/mod_loader.gd
KANAjetzt: Yes - I was thinking about it.
Do we add it to ``mod_loader.gd`` or the setup script?

KANAjetzt:
`--setup-create-override-cfg` is [implemented ](https://github.com/GodotModding/godot-mod-loader/pull/81/commits/8cc7596afb5e5f80786d79f6445bd34d8e2273bf), but for some reason the `override.cfg` is still created.

otDan:
what's missing still @KANAjetzt?

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -32,50 +32,94 @@ const new_global_classes := [
...
+	if is_loader_set_up() and not is_loader_setup_applied():
+		modloaderutils.log_info("ModLoader is set up, the game will be restarted", LOG_NAME)
+		# If the --silent cli argument is passed restart immediately
+		if is_silent:
+			restart_game()
+		# If not start the restart timer and show a notification
+		else:
+			restart_timer.start(4)
+
+		ProjectSettings.set_setting(settings.IS_LOADER_SETUP_APPLIED, true)
```
Qubus0, line 122: this set_setting (and the one in L75) has no effect if it isn't saved to anything.

addons/mod_loader/mod_loader.gd
ithinkandicode: Nvm, I moved all that extra stuff to an issue as it wasn't directly related to this PR, see: #87

This can be resolved now :)

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -90,9 +134,115 @@ func is_loader_setup_applied() -> bool:
...
+	# C:/path/to/game/game.pck
+	path.pck = path.game_base_dir.plus_file(file_name.pck + '.pck')
+	# C:/path/to/game/addons/mod_loader/project.binary
+	path.project_binary = path.mod_loader_dir + "project.binary"
+
+	modloaderutils.log_debug_json_print("path: ", path, LOG_NAME)
+	modloaderutils.log_debug_json_print("file_name: ", file_name, LOG_NAME)
+
+
+func setup_ui() -> void:
```
otDan, line 212: are we sure this will always be on top of everything? 
if for example some autoload has stuff that visualizes a game loading bar or something along those lines

addons/mod_loader/mod_loader_setup.gd
ithinkandicode: Sounds good, as long as CLI args are documented I'm happy :)

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -32,50 +32,94 @@ const new_global_classes := [
...
+	# Checks if the ModLoader Node is in the root of the scene tree
+	# and if the IS_LOADER_SETUP_APPLIED project setting is there
 	if is_loader_setup_applied():
 		modloaderutils.log_info("ModLoader is available, mods can be loaded!", LOG_NAME)
+
 		OS.set_window_title("%s (Modded)" % ProjectSettings.get_setting("application/config/name"))
+
+		var icon := Image.new()
+		var _error_load_icon := icon.load(ProjectSettings.get_setting("application/config/icon"))
+		OS.set_icon(icon)
```
ithinkandicode, line 57: Could you add a brief comment to the code here please, stating why this is needed?

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -32,50 +32,94 @@ const new_global_classes := [
...
-		modloaderutils.log_info("ModLoader is set up, but the game needs to be restarted", LOG_NAME)
-		OS.alert("The Godot ModLoader has been set up. Restart the game to apply the changes. Confirm to quit.")
-		ProjectSettings.set_setting(settings.IS_LOADER_SETUP_APPLIED, true)
-		ProjectSettings.save_custom(modloaderutils.get_override_path())
-		quit()
+	# Add the mods folder if it doesn't exist yet
+	if not modloaderutils.dir_exists(path.game_base_dir + "mods"):
+		var dir = Directory.new()
+		dir.make_dir(path.game_base_dir + "mods")
+		modloaderutils.log_debug("created mods folder" , LOG_NAME)
```
ithinkandicode, line 68: This is great but would it not be better in its own PR?

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -32,50 +32,94 @@ const new_global_classes := [
...
+	if modloaderutils.file_exists(path.pck) and not is_setup_create_override_cfg:
+		modloaderutils.log_debug("injecting the project.binary file", LOG_NAME)
+		create_project_binary()
+		inject_project_binary()
+		clean_up_project_binary_file()
+	# If no dedicated .pck file exists,it's most likely embedded into the .exe.
+	# In that case we add a override.cfg file to the base game dir.
+	else:
+		modloaderutils.log_debug("using the override.cfg file", LOG_NAME)
+		create_override_cfg()
```
ithinkandicode, line 106: This will trigger if either `is_setup_create_override_cfg` is true, or the PCK file doesn't exist. Is this intentional? Or shouldn't it only trigger if `is_setup_create_override_cfg` is true?

It's a nice idea to have things happen *automatically*, but I think it's generally better for things to happen *predictably* -- eg. if a CLI arg exists to perform an action, that action should only be performed if you use that CLI arg.

This would also let you perform both actions, if (for whatever reason) you might want to. Atm this code only lets you create the *override.cfg* if the PCK file doesn't exist, but I don't think we need to add this extra level of conditional complexity. Best to just let users interact with the API without such hidden conditions.

I might be wrong here so I'm open to discussion

addons/mod_loader/mod_loader_setup.gd
ithinkandicode: Maybe make the GUI stuff optional via a CLI arg?

Tbh though I think this GUI feature would be better in a separate subsequent PR, as this PR is focused on patching the binary. Best to keep things small and contained IMO, makes it easier to review, keeps discussions focused, and gets stuff merged faster

ithinkandicode, line 212: Maybe make the GUI stuff optional via a CLI arg?

Tbh though I think this GUI feature would be better in a separate subsequent PR, as this PR is focused on patching the binary. Best to keep things small and contained IMO, makes it easier to review, keeps discussions focused, and gets stuff merged faster

ithinkandicode:
@KANAjetzt Don't forget to fix the merge conflict! ;)

otDan:
poor @KANAjetzt hahaha, we believe in you!

addons/mod_loader/mod_loader_setup.gd
ithinkandicode: I've replied to issue #34 basically saying that I don't think this is needed, eg. for games that use Steam Workshop etc it would just be confusing for us to create this directory

ithinkandicode:
This was closed because it's doing several things, so KANA is going to split it up into smaller chunks, eg:

- #89

PR #89, opened 2023-02-03
Project binary injection

KANAjetzt:
# Project binary injection

Because we cant set the order of autoloads via ``override.cfg`` we have to use external tooling ( [GodotPckTool](https://github.com/hhyyrylainen/GodotPckTool) ) to inject the modified project settings in to the games .pck file.

*Relevant feature proposal:*
https://github.com/godotengine/godot-proposals/discussions/6137

## New functions

- `reorder_autoloads()`
Removes and re adds the autoloads from the `ProjectSettings` to set the `ModLoader` at the top.   

- `create_project_binary()` 
Saves the current `ProjectSettings` to *./addons/mod_loader/project.binary*   

- `inject_project_binary()`
Uses [godotPckTool ](https://github.com/hhyyrylainen/GodotPckTool)to add the custom *project.binary* to the original game pck file.   

- `clean_up_project_binary_file()`
Removes the *project.binary* from *./addons/mod_loader/project.binary*

- `setup_file_data()`
Initializes the 2 dictionary's  `path` and `file_name`

PR #92, opened 2023-02-04
--only-setup cmd arg

KANAjetzt:
*Needs to be merged after [#89 ](https://github.com/GodotModding/godot-mod-loader/pull/89)*

Adds cmd arg:

-  `--only-setup`
Allows to setup the mod loader without user notification, or auto restart ( upcoming PR ).

Request from Mythic ( Thunderstore ):
> Mythic — 30/01/2023 09:39
how about would it be possible to support running just the installation step without restarting, but rather shutting down once it's complete?
that way we can use that as the "installer" on the manager, and you don't need to worry about CLI arg handling in the restart phase

> Mythic — 30/01/2023 09:43
and if that can return an appropriate exit code, (0 for success, non-zero for fail), that'd be great too
then we can have the manager just run that for install


KANAjetzt:
I think I messed that up

PR #95, opened 2023-02-04
Fix global class registration

Qubus0:
fix: prevent duplicate global class registration
refactor: utils global class registration does not save anymore, mod_loader exposes a new method for mods to use it and save

### DIFF
--- a/addons/mod_loader/mod_loader_utils.gd
+++ b/addons/mod_loader/mod_loader_utils.gd
@@ -260,7 +260,7 @@ static func register_global_classes_from_array(new_global_classes: Array) -> voi
 				if OS.has_feature("editor"):
 					log_info('Class "%s" to be registered as global was already registered by the editor. Skipping.' % new_class.class, LOG_NAME)
 				else:
-					log_warning('Class "%s" to be registered as global already exists. Skipping.' % new_class.class, LOG_NAME)
+					log_info('Class "%s" to be registered as global already exists. Skipping.' % new_class.class, LOG_NAME)
 				continue
 
 		registered_classes.append(new_class)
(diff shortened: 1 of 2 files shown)


addons/mod_loader/mod_loader_utils.gd
```diff
@@ -248,25 +248,26 @@ static func dir_exists(path: String) -> bool:
...
 		if not is_valid_global_class_dict(new_class):
 			continue
-		if registered_classes.has(new_class):
-			continue
+		for old_class in registered_classes:
+			if old_class.class == new_class.class:
+				if OS.has_feature("editor"):
+					log_info('Class "%s" to be registered as global was already registered by the editor. Skipping.' % new_class.class, LOG_NAME)
+				else:
+					log_warning('Class "%s" to be registered as global already exists. Skipping.' % new_class.class, LOG_NAME)
```
KANAjetzt, line 263: Maybe also ``log_info()`` here?
Warning sounds like something you should fix.

Qubus0: I think it is something you should fix. Either you have the same thing twice, or another mod uses that class name - in which case you need to change yours

KANAjetzt: If I call `register_global_classes_from_array(new_global_classes: Array)` from a mod.
That class will be added each time the mod is loaded.
I'm not sure the mod author can do something to prevent that?

Qubus0: good point, changed!

KANAjetzt (approved):
ty 👍

PR #100, opened 2023-02-06
store log backups and rotate them like godot

### LINKED ISSUE #16: save logs up until limit before overriding the files like Godot does it
to prevent logs being overridden immediately, I propose that we keep 4 (or what is set in ProjectSettings `logging/file_logging/max_log_files`) older logs, with the time they were created in the name

Qubus0:
Every game start now gets their own log and they are kept up to the same limit set for the vanilla game (5 usually)

- closes #16


KANAjetzt (approved):
I can't test it right now - but everything looks good. 👍

PR #103, opened 2023-02-08
fix: properly parse all types of command line arg values

Qubus0:
This can now properly account for args in format `--arg="value"` and also `--arg "value"` (like `--script`)
It also reverses the Godot bug present in editor builds for Linux and Mac where it would split quoted args at whitespace in this format `--arg="value with whitespace"`
The last thing is that `is_running_with_command_line_arg` will now also return true for args in this format `--arg="value"`, which it didn't before since it had to be an exact match

testing: just put this into the mod loader _init

```gdscript
func _init() -> void:
	var test_args := ["--script", "--fake-arg", "--test-mods-path", "--log-info", "--log-ignore"]
	for test_arg in test_args:
		printt(
			"running with %s?" % test_arg,
			ModLoaderUtils.is_running_with_command_line_arg(test_arg),
			ModLoaderUtils.get_cmd_line_arg_value(test_arg)
		)
```
I used these args 
`--script addons/mod_loader/mod_loader_setup.gd --log-info --log-ignore=ModLoader:ModData --test-mods-path="C :\s ome\path wit h\whi te space\"`

### DIFF
--- a/addons/mod_loader/mod_loader_utils.gd
+++ b/addons/mod_loader/mod_loader_utils.gd
@@ -222,10 +222,9 @@ static func get_fixed_cmdline_args() -> PoolStringArray:
 
 
 # Reverses a bug in Godot, which splits input strings at spaces even if they are quoted
-# e.g. --arg="some value" becomes [ --arg="some, value" ]
-# this only occurs with args in that format. --arg "some value" is safe
+# e.g. `--arg="some value" --arg-two 'more value'` becomes `[ --arg="some, value", --arg-two, 'more, value' ]`
 static func fix_godot_cmdline_args_string_space_splitting(args: PoolStringArray) -> PoolStringArray:
-	if not OS.has_feature("editor"): # only appears in editor builds
+	if not OS.has_feature("editor"): # only happens in editor builds
 		return args
 	if OS.has_feature("Windows"): # windows is unaffected
 		return args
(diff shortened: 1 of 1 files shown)


addons/mod_loader/mod_loader_utils.gd
```diff
@@ -188,25 +188,70 @@ static func _get_verbosity() -> int:
...
 # Get the command line argument value if present when launching the game
 static func get_cmd_line_arg_value(argument: String) -> String:
-	for arg in OS.get_cmdline_args():
-		if (arg as String).find("=") > -1:
-			var key_value := (arg as String).split("=")
-			# True if the checked argument matches a user-specified arg key
-			# (eg. checking `--mods-path` will match with `--mods-path="C://mods"`
-			if key_value[0] == argument:
-				return key_value[1]
+	var args := fix_godot_cmdline_args_string_space_splitting(OS.get_cmdline_args())
```
ithinkandicode, line 199: Would it be useful to also add a wrapper func for this, eg `get_fixed_cmdline_args`?

addons/mod_loader/mod_loader_utils.gd
```diff
@@ -188,25 +188,70 @@ static func _get_verbosity() -> int:
...
+
+	for arg_index in args.size():
+		var arg := args[arg_index] as String
+
+		var key := arg.split("=")[0]
+		if key == argument:
+			# format: [code]--arg=value[/code] or [code]--arg="value"[/code]
```
ithinkandicode, line 206: I can see that using `[code]` instead of ``` `code` ``` is part of the [Godot 4 docs](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_documentation_comments.html), but we've used ``` `code` ``` throughout the rest of the files, so perhaps it would be better to maintain that standard for new additions?

And we could discuss potentially converting the existing doc comments to the Godot 4 standard in a subsequent issue/PR

ithinkandicode (approved):
👍

PR #99, opened 2023-02-06
--setup-create-override-cfg cli arg

KANAjetzt:
Adds cmd arg:

###  --setup-create-override-cfg
Forces the setup to create the ``override.cfg`` in the game base directory.
Skips the ``project.binary`` injection.

   
   
The project binary injection is skipped in that case.
```python
	if is_setup_create_override_cfg:
		handle_override_cfg()
	else:
		handle_project_binary()
```

PR #104, opened 2023-02-08
Utils for ModLoader autoload position

KANAjetzt:
Adds two functions to ModLoaderUtils so they can be used in `mod_loader.gd` and the setup script.

- `get_autoload_array() -> Array`
Get an array of all autoloads -> ["autoload/AutoloadName", ...]

- `get_autoload_index(autoload_name: String) -> int`
Get the index of a specific autoload

Updates code from #96 to use the new util.

Has to be merged after #99

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -169,7 +169,7 @@ func _init() -> void:
 # Ensure ModLoader is the first autoload
 func _check_first_autoload() -> void:
 	var autoload_array = ModLoaderUtils.get_autoload_array()
-	var is_mod_loader_first = autoload_array.find("autoload/ModLoader") == 0
+	var is_mod_loader_first = autoload_array.find("ModLoader") == 0
 
 	var base_msg = "ModLoader needs to be the first autoload to work correctly, "
 	var help_msg = ""
@@ -183,7 +183,7 @@ func _check_first_autoload() -> void:
 		help_msg = "If you're seeing this error, something must have gone wrong in the setup process."
 
 	if not is_mod_loader_first:
-		ModLoaderUtils.log_fatal(str(base_msg, 'but the first autoload is currently: "%s". ' % autoload_array[0].trim_prefix("autoload/"), help_msg), LOG_NAME)
+		ModLoaderUtils.log_fatal(str(base_msg, 'but the first autoload is currently: "%s". ' % autoload_array[0], help_msg), LOG_NAME)
 
 	if autoload_array.size() == 0:
 		ModLoaderUtils.log_fatal(str(base_msg, "but no autoloads are currently set up. ", help_msg), LOG_NAME)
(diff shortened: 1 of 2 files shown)


addons/mod_loader/mod_loader_utils.gd
```diff
@@ -350,6 +350,30 @@ static func is_valid_global_class_dict(global_class_dict: Dictionary) -> bool:
 	return true
 
 
+# Get an array of all autoloads -> ["autoload/AutoloadName", ...]
+static func get_autoload_array() -> Array:
+	var autoloads := {}
```
Qubus0, line 355: might as well just use an array here and append each name instead of the key value stuff. you can also use String`.trim_prefix("autoload/")` to only put the name in the array, then  you don't have to do this `autoload_array[0].split('/')[1]` to get the name or add it when searching for the index

KANAjetzt: ty as always 👍 

*fixed in:
6a2c22b
f6181cf*

Qubus0 (approved):
Perfect 👍

addons/mod_loader/mod_loader.gd
```diff
@@ -168,35 +168,24 @@ func _init() -> void:
...
-				if name == "autoload/ModLoader":
-					is_mod_loader_first = true
-			var value: String = ProjectSettings.get_setting(name)
-			autoloads[name] = value
-			autoload_index += 1
-
-	# Log the autoloads order. Might seem superflous but could help when providing support
-	ModLoaderUtils.log_debug_json_print("Autoload order", autoloads, LOG_NAME)
+	var autoload_array = ModLoaderUtils.get_autoload_array()
+	var is_mod_loader_first = autoload_array.find("ModLoader") == 0
```
ithinkandicode, line 172: Love this, the new utility funcs make it so much cleaner and let us re-use the code! ❤

addons/mod_loader/mod_loader.gd
```diff
@@ -168,35 +168,24 @@ func _init() -> void:
...
+	ModLoaderUtils.log_debug_json_print("Autoload order", autoload_array, LOG_NAME)
+
 	if OS.has_feature("editor"):
 		help_msg = "To configure your autoloads, to go Project > Project Settings > Autoload, and add ModLoader as the first item. For more info, see the 'Godot Project Setup' page on the ModLoader GitHub wiki."
 	else:
 		help_msg = "If you're seeing this error, something must have gone wrong in the setup process."
 
 	if not is_mod_loader_first:
-		ModLoaderUtils.log_fatal(str(base_msg, 'but the first autoload is currently: "%s". ' % name, help_msg), LOG_NAME)
+		ModLoaderUtils.log_fatal(str(base_msg, 'but the first autoload is currently: "%s". ' % autoload_array[0], help_msg), LOG_NAME)
```
ithinkandicode, line 186: clever ;)

ithinkandicode (approved):
yep yep love it, great work KANA 👍

PR #110, opened 2023-02-13
✔ added checks for override.cfg setup

KANAjetzt:
## mod_loader_setup.gd

Prevents the setup script from running the setup again, if the mod loader was installed via the *override.cfg*.

 ## mod_loader.gd

Prevents the error logging in `_check_first_autoload()`, if a *override.cfg* file is present.

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -169,16 +169,26 @@ func _init() -> void:
 # Ensure ModLoader is the first autoload
 func _check_first_autoload() -> void:
 	var autoload_array = ModLoaderUtils.get_autoload_array()
-	var is_mod_loader_first = autoload_array.find("ModLoader") == 0
+	var mod_loader_index = autoload_array.find("ModLoader")
+	var is_mod_loader_first = mod_loader_index == 0
 
-	var base_msg = "ModLoader needs to be the first autoload to work correctly, "
-	var help_msg = ""
+	var override_cfg_path = ModLoaderUtils.get_override_path()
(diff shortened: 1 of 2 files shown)


KANAjetzt:
I think this code will never run, if there is no autoloads there will be no mod_loader.gd right?

https://github.com/GodotModding/godot-mod-loader/blob/ae19b71bec47fd02135f47251f3d4a4060efba29/addons/mod_loader/mod_loader.gd#L198-L199

Qubus0:
you are right, that will never run. the mod loader will always occupy one slot or not be around to tell anyone. this should be in the setup, if at all.

addons/mod_loader/mod_loader.gd
```diff
@@ -169,16 +169,26 @@ func _init() -> void:
...
 	# Log the autoloads order. Might seem superflous but could help when providing support
 	ModLoaderUtils.log_debug_json_print("Autoload order", autoload_array, LOG_NAME)
 
+	# If the override file exists we assume the ModLoader was setup with the --setup-create-override-cfg cli arg
+	# In that case the ModLoader will be the last entry in the autoload array
+	if is_override_cfg_setup:
+		ModLoaderUtils.log_info("override.cfg setup detected, ModLoader will be the last autoload loaded.", LOG_NAME)
+		return
+
+	var base_msg = "ModLoader needs to be the first autoload to work correctly, "
```
Qubus0, line 187: More like "ModLoader needs to be the first autoload to work correctly if you are using mods that extend any autoloads"

Qubus0 (approved):
looks good

PR #121, opened 2023-02-18
Extensions with inheritance

GDami:
The current schema for installing script extensions relies on an _importance_ score for each mod, where the score is equal to how many times a mod appears as a dependency inside the current mod list.
That score is used to let more important mods load before others, and thus let their extensions load before those of other mods.

This layout poses one problem in situations where parent and child scripts in vanilla are supposed to be extended within the same mod list : a mod with bigger importance can look to extend a script that is child to a script extended by mod(s) of lower importance. 

This results in the child script not taking into account the changes made by the extension(s) to the parent script.
This is also not a problem that can be solved through mod load order alone, since regardless of the importance score, 2 mods may be faced with this situation twice, with their roles reversed in the second situation.

This PR looks to take into account vanilla's inheritance chain to decide the order in which extensions are installed. The mods' importance scores are still used to load mods following their dependencies.
But in situations where a more important mod would extend a child before their parent, the extension(s) of the parent script is given priority.

Also, even when extending in that order, vanilla parents still get loaded as base scripts when extending a child script, instead of their extended version.
To make sure the extended parents are used when extending child scripts, reload() is called everytime a script is extended


addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +476,171 @@ func _init_mod(mod: ModData) -> void:
...
+	
+	
+	return extensions_sorted
+
+
+# goes up a's inheritance tree to find if any parent shares the same path as b
+func check_inheritances(a:Array, b:Array)->bool:
+	var a_child_script
+	if loaded_vanilla_parents.keys().has(a[1]):
+		a_child_script = ResourceLoader.load(a[1])
```
Qubus0, line 591: the readability here would benefit a lot from an internal class
```gdscript
func check_inheritances(a:Array, b:Array)->bool:          # <- some breathing room makes these easier to read
	var a_child_script
	if loaded_vanilla_parents.keys().has(a[1]):
		a_child_script = ResourceLoader.load(a[1])
    # ...
```
```gdscript
func check_inheritances(extension_a: ScripExtensionData, extension_a: ScripExtensionData) -> bool:
	var extension_a_child_script: Resource    # could even be Script, didn't test that yet
	if loaded_vanilla_parents.keys().has(extension_a.parent_script_path):
		extension_a_child_script = ResourceLoader.load(extension_a.parent_script_path)
    # ...
```

addons/mod_loader/mod_loader.gd
```diff
@@ -489,11 +649,15 @@ func install_script_extension(child_script_path: String) -> void:
 	# when creating singletons.
 	# The actual instance is thrown away.
 	child_script.new()
-
+
```
Qubus0, line 652: you can activate the editor setting `text_editor/files/trim_trailing_whitespace_on_save` to avoid these

Qubus0 (commented):
Thanks a lot for your PR 😃 
I have a few code style things that you should apply throughout your code to make things easier to understand and also to protect against bugs in the future.
But otherwise: well done!

GDami:
I went ahead and applied your suggestions, code should look better now !

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +516,125 @@ func _init_mod(mod: ModData) -> void:
...
 # Note that your extender script doesn't have to follow the same directory path
 # as the vanilla file, but it's good practice to do so.
-func install_script_extension(child_script_path: String) -> void:
-	# Check path to file exists
-	if not File.new().file_exists(child_script_path):
-		ModLoaderUtils.log_error("The child script path '%s' does not exist" % [child_script_path], LOG_NAME)
-		return
+func install_script_extension(child_script_path:String):
+	
+	script_extensions.push_back(child_script_path)
```
ithinkandicode, line 521: Does this mean that `install_script_extension` can't install extensions after ModLoader has finished its setup? [Here's](https://github.com/BrotatoMods/Brotato-BFX/blob/main/root/mods-unpacked/Darkly77-BFX/extensions/singletons/debug_service.gd) an example use case. 

Would it be possible to still support this functionality? Eg. with a func called `install_late_script_extension` or something. Or is this something we no longer want to support?

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +516,125 @@ func _init_mod(mod: ModData) -> void:
...
+
+
+func handle_script_extensions()->void:
+	
+	# couple the extension paths with the parent paths and the extension's mod id
```
ithinkandicode, line 526: Hate to be fussy but can the first letter be capitalised please, so it matches the standard of other comments?

addons/mod_loader/script_extension_data.gd
```diff
@@ -0,0 +1,16 @@
+class_name ScriptExtensionData
+extends Resource
+
+
```
ithinkandicode, line 4: Could you add a comment with a brief description of the purpose of this file? It can help other developers better understand the code without needing to dig so deeply. Eg see [mod_data.gd](https://github.com/GodotModding/godot-mod-loader/blob/main/addons/mod_loader/mod_data.gd) and [mod_manifest.gd](https://github.com/GodotModding/godot-mod-loader/blob/main/addons/mod_loader/mod_manifest.gd)

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +516,125 @@ func _init_mod(mod: ModData) -> void:
...
+			ScriptExtensionData.new(extension_path, parent_script_path, mod_id)
+		)
+	
+	# Sort the extensions based on dependencies
+	parent_paths = sort_extensions_from_load_order(parent_paths)
+	# Inheritance is more important so this called last
+	parent_paths.sort_custom(self, "check_inheritances")
+	
+	# Maybe useless, i think this saved some bugs in the past. 
+	loaded_vanilla_parents_cache.clear()
```
ithinkandicode, line 555: Could the comment here be re-written to exclude the first-person?

ithinkandicode (commented):
Looks good, just some minor comments.

Also, I'm not sure I fully understand the purpose of this. From Discord you said that it can fix the issue of not being able to extend certain files, which is huge! But the current PR description doesn't tell me an awful lot about things. Ideally it would be good to know:

a) What the current problem is, and why that's bad/limiting/etc
b) How this PR solves it (tho I think you've already covered this)

_edit:_ The PR description has been updated now, thank you 🤗

addons/mod_loader/mod_loader.gd
Qubus0: oh that is a good point. we should not lose that

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +481,128 @@ func _init_mod(mod: ModData) -> void:
...
+	else:
+		apply_extension(child_script_path)
+
+
+func handle_script_extensions()->void:
+	
+	# Couple the extension paths with the parent paths and the extension's mod id
+	# in a ScriptExtensionData resource
+	
+	var script_extension_datas := []
```
Qubus0, line 507: `script_extension_data_array` might be better

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +481,128 @@ func _init_mod(mod: ModData) -> void:
...
+	for extension_path in script_extensions:
+		
+		if not File.new().file_exists(extension_path):
+			ModLoaderUtils.log_error("The child script path '%s' does not exist" % [extension_path], LOG_NAME)
+			continue 
+		
+		var child_script = ResourceLoader.load(extension_path)
+		
+		var mod_id:String = extension_path.replace(UNPACKED_DIR, "").get_slice("/", 0)
```
Qubus0, line 516: you could also `extension_path.trim_prefix(UNPACKED_DIR).get_slice("/", 0)`

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +487,128 @@ func _init_mod(mod: ModData) -> void:
...
+	
+	# Inheritance is more important so this called last
+	script_extension_datas.sort_custom(self, "check_inheritances")
+	
+	# This saved some bugs in the past. 
+	loaded_vanilla_parents_cache.clear()
+	
+	# Load and install all extensions
+	for extension in script_extension_datas:
+		var script:Script = apply_extension(extension["extension_path"])
```
Qubus0, line 539: you should be able to access this with `extension.extension_path`

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +487,128 @@ func _init_mod(mod: ModData) -> void:
...
+
+
+func sort_extensions_from_load_order(extensions:Array)->Array:
+	
+	# Populate extensions_sorted following the dependency order
+	var extensions_sorted := []
+	
+	for _mod_data in mod_load_order:
+		for script in extensions:
+			if script["mod_id"] == _mod_data.dir_name:
```
Qubus0, line 550: this one as well

Qubus0 (commented):
getting there!

addons/mod_loader/mod_loader.gd
GDami: oh i didn't know this one, cool

addons/mod_loader/mod_loader.gd
```diff
@@ -473,14 +487,128 @@ func _init_mod(mod: ModData) -> void:
...
+	# extensions to be installed taking inheritance chain into account
+	if is_initializing:
+		script_extensions.push_back(child_script_path)
+	
+	# If not, apply the extension directly
+	else:
+		apply_extension(child_script_path)
+
+
+func handle_script_extensions()->void:
```
ithinkandicode, line 502: Re: This method and all the ones below it, it looks like they should be private methods (ie. prefixed with `_`) that are only used internally by ModLoader within this script.

Atm they're in the "Helpers" methods group and they aren't prefixed with `_`, which means that they're intended to be API methods for modders to use.

GDami: I went ahead and moved them out of the Helpers methods.

addons/mod_loader/mod_loader.gd
ithinkandicode: @GDami Just this tiny whitespace issue to fix then I think we're all good! \o/

GDami: The white space is gone 👍

ithinkandicode (approved):
Amazing work, thank you dami 🤗

I've tested this on my local build of Brotato, with all my usual mods, and everything works fine 👍

KANAjetzt:
closed in favor of #124

PR #128, opened 2023-02-24
refactor: ♻️ Added is_silent parameter to is_name_or_namespace_valid()

KANAjetzt:
`is_silent` parameter for `is_name_or_namespace_valid()`. Can be used to not log an error if the validation returns `false`. Useful for example if the validation is used in a text input where the result will be `false`, most of the times.

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -177,6 +177,8 @@ func _init() -> void:
 	
 	_handle_script_extensions()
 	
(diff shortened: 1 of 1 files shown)


ithinkandicode:
Is this for the dependency version issue? If so, please see:

- #129

Qubus0 (commented):
good, but we need those for other validation methods as well
@ithinkandicode yeah it's for the addon, to not spam the console with each validation. Probably a good thing to have anyway though, even if it's technically not needed in the mod loader itself yet.

ithinkandicode (approved):
👍

PR #130, opened 2023-02-25
refactor: :recycle: Added is_silent parameter to validation functions

KANAjetzt:
Added is_silent parameter to 

- `is_name_or_namespace_valid()`
- `is_semver_valid()`
- `validate_dependencies_and_incompatibilities()`
- `is_mod_id_valid()`

Can be used to not log an error if the validation returns `false`. Useful for example if the validation is used in a text input where the result will be most of the times `false`. Implemented with the [ModTool](https://github.com/GodotModding/godot-mod-tool) in mind.

**Sorry accidentally closed #128*


Qubus0 (approved):
perfect, thanks

PR #163, opened 2023-03-02
refactor: :recycle: ModManifest - Added type checks

KANAjetzt:
Added type check to `_get_string_from_dict()` and  `_get_array_from_dict()`


Qubus0 (approved):
how could I miss that lol. We could also return str(dict[key]), but I'm not sure if that would be considered the "expected" behaviour

PR #175, opened 2023-03-08
feat: :sparkles: ModManifest - Added `to_json()` function

KANAjetzt:
Can be used to retrieve the Manifest values as JSON, in the *manifest.json* format. Will be used in the Mod Tool.


Qubus0 (approved):
Simple, nice

PR #177, opened 2023-03-11
style: :art: Add `\t` indent to `JSON.print()`

KANAjetzt:
This will make the generated JSON a lot more readable if it is saved to a file.


ithinkandicode (approved):
Hell yeah :D

PR #183, opened 2023-03-24
fix: :bug: circular dependency check

### LINKED ISSUE #111: Circular dependency - infinite loop in dependency check
If 2 mods depend on each other the dependency check will run into infinity.

KANAjetzt:
Before this fix, it was possible to trigger an infinite loop in the dependency check if two mods declared each other as dependencies. Now, the dependency chain is tracked inside the `dependency_chain` array. If the same mod ID is detected, the loop will break.

closes #111

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -430,7 +430,7 @@ func _check_dependencies(mod: ModData, dependency_chain := []) -> bool:
 	var mod_id := mod.dir_name
 
 	# Check for circular dependency
-	if dependency_chain.find(mod_id) != -1:
+	if mod_id in dependency_chain:
 		is_circular = true
 		ModLoaderUtils.log_debug("Dependency check - circular dependency detected.", LOG_NAME)
 		return is_circular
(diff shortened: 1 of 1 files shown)


addons/mod_loader/mod_loader.gd
```diff
@@ -421,14 +423,26 @@ func _init_mod_data(mod_folder_path: String) -> void:
...
 # hasn't been loaded, the dependent mod won't be loaded.
-func _check_dependencies(mod: ModData) -> void:
+func _check_dependencies(mod: ModData, dependency_chain := []) -> bool:
 	ModLoaderUtils.log_debug("Checking dependencies - mod_id: %s dependencies: %s" % [mod.dir_name, mod.manifest.dependencies], LOG_NAME)
 
+	var is_circular := false
+	var mod_id := mod.dir_name
+
+	# Check for circular dependency
+	if dependency_chain.find(mod_id) != -1:
```
Qubus0, line 433: you could use the `in` keyword. `if mod_id in dependency_chain:`

Qubus0 (approved):
lookin' good

addons/mod_loader/mod_loader.gd
KANAjetzt: *fixed [d6707dd](https://github.com/GodotModding/godot-mod-loader/pull/183/commits/d6707dd230b4247a26e5930c1f60947853e557d4)*

PR #184, opened 2023-03-24
fix: :bug: continue check if circular dep is detected

KANAjetzt:
Instead of returning if a circular dependency is detected, the loop should continue with the next entry.


Qubus0 (approved):
makes sense

PR #115, opened 2023-02-15
✔ Load before

KANAjetzt:
# Load before some mod

Adds functionality to specify a mod you want to load before your own mod. 

Load before field added to  *manifest.json* in *extra* -> *godot* -> *load_before*

For example:
`"load_before": ["Darkly77-ContentLoader"],`

```JSON
{
	"name": "WhatAmILookingAt",
	"namespace": "KANA",
	"version_number": "1.0.0",
	"description": "Tells you where this thing is from.",
	"website_url": "https://github.com/BrotatoMods",
	"dependencies": [],
	"extra": {
		"godot": {
			"incompatibilities": [],
			"load_before": ["Darkly77-ContentLoader"],
			"authors": ["KANA"],
			"compatible_mod_loader_version": "4.3.0",
			"compatible_game_version": ["0.8.0.3"],
			"config_defaults": {}
		}
	}
}
```

Will need some more testing, this is just a naive implementation.

---
closes #114

PR #185, opened 2023-03-24
feat: :sparkles: added load_before

### LINKED ISSUE #114: Add load before functionality
>KANA — Yesterday at 16:46
 I need a way to tell a mod ( Darkly77-ContentLoader ) that it has to wait for my mod ( KANA-WhatAmILookingAt )

> Ste — Yesterday at 16:49
technically only by dependencies.. we could introduce a "load-before" to store more mod ids that should load after this mod
load before would essentially copy over the importance score of the other mod and add it to the own importance

KANAjetzt:
The code from 
- #115 

## *Original Description*

Adds functionality to specify a mod you want to load before your own mod. 

Load before field added to  *manifest.json* in *extra* -> *godot* -> *load_before*

For example:
`"load_before": ["Darkly77-ContentLoader"],`

```JSON
{
	"name": "WhatAmILookingAt",
	"namespace": "KANA",
	"version_number": "1.0.0",
	"description": "Tells you where this thing is from.",
	"website_url": "https://github.com/BrotatoMods",
	"dependencies": [],
	"extra": {
		"godot": {
			"incompatibilities": [],
			"load_before": ["Darkly77-ContentLoader"],
			"authors": ["KANA"],
			"compatible_mod_loader_version": "4.3.0",
			"compatible_game_version": ["0.8.0.3"],
			"config_defaults": {}
		}
	}
}
```

Will need some more testing, this is just a naive implementation.

---
closes #114 

---

## I tested it a bit more

- Mods added to `load_before` should not be added to `dependencies` this cancels out the effect.
Because a cyclic reference is created and the dependency check will break out of the loop, so both mods end up with the same importance score.
- It worked successfully for this simple test case
   - Mod 1 no dependencies
   - Mod 2 `load_before` Mod 1
   - Mod 3 `dependency` on Mod 1

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -493,13 +493,24 @@ func _check_load_before(mod: ModData) -> void:
 	if mod.manifest.load_before.size() == 0:
 		return
 
-	# Add the mod to the dependency arrays
+	# For each mod id in load_before
 	for load_before_id in mod.manifest.load_before:
 		var load_before_mod_dependencies = mod_data[load_before_id].manifest.dependencies
+		# Check if it's already a dependency
+		if mod.dir_name in load_before_mod_dependencies:
+			ModLoaderUtils.log_debug(
+				"Load before detected -> Skipping %s because it's already a dependency for %s"
+				% [mod.dir_name, load_before_id], LOG_NAME
+			)
+			return
+		# Add the mod to the dependency array
 		load_before_mod_dependencies.append(mod.dir_name)
 		mod_data[load_before_id].manifest.dependencies = load_before_mod_dependencies
 
-	ModLoaderUtils.log_debug("Load before detected -> Added %s as dependency for %s" % [mod.dir_name, mod.manifest.load_before], LOG_NAME)
+	ModLoaderUtils.log_debug(
+		"Load before detected -> Added %s as dependency for %s"
+		% [mod.dir_name, mod.manifest.load_before], LOG_NAME
+	)
 
 
 # Get the load order of mods, using a custom sorter
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -146,9 +146,9 @@ func _init() -> void:
 		var mod: ModData = mod_data[dir_name]
 		if not mod.is_loadable:
 			continue
-		var is_circular := _check_dependencies(mod)
-		if is_circular:
-			return
+			var is_circular := _check_dependencies(mod)
+			if is_circular:
+				return
 
 	# Sort mod_load_order by the importance score of the mod
 	mod_load_order = _get_load_order(mod_data.values())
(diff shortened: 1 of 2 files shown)


addons/mod_loader/mod_loader.gd
```diff
@@ -474,6 +485,23 @@ func _handle_missing_dependency(mod_dir_name: String, dependency_id: String) ->
...
+func _check_load_before(mod: ModData) -> void:
+	# Skip if no entries in load_before
+	if mod.manifest.load_before.size() == 0:
+		return
+
+	# Add the mod to the dependency arrays
+	for load_before_id in mod.manifest.load_before:
+		var load_before_mod_dependencies = mod_data[load_before_id].manifest.dependencies
+		load_before_mod_dependencies.append(mod.dir_name)
+		mod_data[load_before_id].manifest.dependencies = load_before_mod_dependencies
```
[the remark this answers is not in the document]
KANAjetzt: dependencies is a `PoolStringArray` and they like to be reassigned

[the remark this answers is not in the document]
Qubus0: oh I see

addons/mod_loader/mod_loader.gd
```diff
@@ -474,6 +485,23 @@ func _handle_missing_dependency(mod_dir_name: String, dependency_id: String) ->
...
+# mod_manifest (ie. its manifest.json file). Add the mod to the dependency of the
+# mods inside the load_before array.
+func _check_load_before(mod: ModData) -> void:
+	# Skip if no entries in load_before
+	if mod.manifest.load_before.size() == 0:
+		return
+
+	# Add the mod to the dependency arrays
+	for load_before_id in mod.manifest.load_before:
+		var load_before_mod_dependencies = mod_data[load_before_id].manifest.dependencies
```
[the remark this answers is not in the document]
KANAjetzt: ty - *fixed [3266fde](https://github.com/GodotModding/godot-mod-loader/pull/185/commits/3266fde01f64a742ed08e8649db4bad490164a89)*

addons/mod_loader/mod_loader.gd
```diff
@@ -493,13 +493,24 @@ func _check_load_before(mod: ModData) -> void:
 	if mod.manifest.load_before.size() == 0:
 		return
 
-	# Add the mod to the dependency arrays
+	# For each mod id in load_before
 	for load_before_id in mod.manifest.load_before:
 		var load_before_mod_dependencies = mod_data[load_before_id].manifest.dependencies
+		# Check if it's already a dependency
+		if mod.dir_name in load_before_mod_dependencies:
```
Qubus0, line 500: oh this is a good catch. what I originally meant was, that it should also check if the load before mod is actually part of the mods that are available and enabled, so that there won't be any issues with nonexistent dependencies because they were added as "load before"

KANAjetzt: oh right - good point ty 👍

KANAjetzt: is this handled up here?

https://github.com/GodotModding/godot-mod-loader/blob/3266fde01f64a742ed08e8649db4bad490164a89/addons/mod_loader/mod_loader.gd#L136-L140

Qubus0: no, that only addresses tho mod that has the `load_before` array. it needs to check if a mod from the `load_before` array is in `mod_data`

Qubus0: Wait, load before inserts itself into the dependencies of the other mod. I had confused myself.

On another note though - there needs to be a check if that other mod exists. Before line 498

KANAjetzt: Good catch 👍 
*fixed in [c6f3a75](https://github.com/GodotModding/godot-mod-loader/pull/185/commits/c6f3a758e609dc5de16d2af0af88aebc6d350211)*

ithinkandicode (commented):
Looks good to me, only one suggestion: The PR description says that a mod shouldn't list another mod in both its `load_before` and `dependencies`, so could there be an error if a mod developer tries to do this?

Qubus0 (approved):
nice, well done.
@ithinkandicode duplicate in deps and load before should be handled here, that remark is just outdated
https://github.com/GodotModding/godot-mod-loader/blob/ff8e9d85dac1d6d66b27ebbfc809c409c5a8a311/addons/mod_loader/mod_loader.gd#L508-L511

KANAjetzt:
@ithinkandicode I opened #187 for that.
I will fix this with a manifest validation 👍
Ty for the review :)

PR #193, opened 2023-03-28
refactor: :recycle: `validate_distinct_mod_ids_in_arrays()`

### LINKED ISSUE #192: `incompatibility` should not be an `optional_dependency`
> nice, thanks. not sure if we need it but technically, an incompatibility should also not be an optional dependency. might not be in the scope here since the optional ones were added after this issue

_Originally posted by @Qubus0 in https://github.com/GodotModding/godot-mod-loader/pull/174#pullrequestreview-1360769790_

KANAjetzt:
- reworked `validate_dependencies_and_incompatibilities_conflicts()` to work with any arrays
-  renamed it to `validate_distinct_mod_ids_in_arrays()`. 
- Fixed the accidentally removed function call of `validate_dependencies_and_incompatibilities_conflicts()` in `_init()`
- Added validation to check for distinct mod_ids in `incompatibilities` and `optional_dependencies`.

closes #192

### DIFF
--- a/addons/mod_loader/classes/mod_manifest.gd
+++ b/addons/mod_loader/classes/mod_manifest.gd
@@ -266,6 +266,7 @@ static func validate_distinct_mod_ids_in_arrays(
 	array_one: PoolStringArray,
 	array_two: PoolStringArray,
 	array_description: PoolStringArray,
+	additional_info := "",
 	is_silent := false
 ) -> bool:
 	# Initialize an empty array to hold any overlaps.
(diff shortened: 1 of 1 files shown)


addons/mod_loader/classes/mod_manifest.gd
```diff
@@ -249,35 +261,36 @@ static func is_semver_valid(check_version_number: String, is_silent := false) ->
...
+					% [mod_id, overlaps, array_description[0], array_description[1], additional_info]
 				),
 				LOG_NAME
 			)
-		return true
+		return false
 
 	# If no overlaps were found, return false.
-	return false
+	return true
```
Qubus0, line 293: comment not accurate anymore.
might be more self explanatory to just use an early return earlier

```gdscript
if overlaps.size() == 0:
    return true
if not is_silent:
    ...
return false
```

Qubus0 (approved):
👍

addons/mod_loader/classes/mod_manifest.gd
KANAjetzt: *fix 2 in https://github.com/GodotModding/godot-mod-loader/pull/193/commits/154f1ca605df2784d3a98247a1fa861d0363ed4b* 👀

Qubus0 (approved):
nice

PR #195, opened 2023-03-29
feat: added _reset_extension

otDan:
Adds `_reset_extension()` so that an extension can be reset to the default state of the base godot project script, we save the base scripts when extending one, all the scripts saved are saved in order of calling, this is the base for the subsequent prs that will allow for easy mod updating and disabling at runtime

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -721,7 +721,8 @@ func _apply_extension(extension_path)->Script:
 	return child_script
 
 
-func _reset_extension(parent_script_path)->Script:
+# Used to fully reset the provided script to a state prior of any extension
+func _reset_extension(parent_script_path: String)->Script:
 	# Check path to file exists
 	if not File.new().file_exists(parent_script_path):
 		ModLoaderUtils.log_error("The parent script path '%s' does not exist" % [parent_script_path], LOG_NAME)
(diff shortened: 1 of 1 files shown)


addons/mod_loader/mod_loader.gd
```diff
@@ -703,12 +705,44 @@ func _apply_extension(extension_path)->Script:
...
+	_saved_scripts[parent_script_path].append(child_script)
+	ModLoaderUtils.log_error("base script: %s" % str(_saved_scripts[parent_script_path]), LOG_NAME)
+
 	ModLoaderUtils.log_info("Installing script extension: %s <- %s" % [parent_script_path, extension_path], LOG_NAME)
 	child_script.take_over_path(parent_script_path)
 
 	return child_script
 
 
+func _reset_extension(parent_script_path)->Script:
```
KANAjetzt, line 724: `func _reset_extension(parent_script_path)->Script:`

Type for parameter and some white space.
`func _reset_extension(parent_script_path: String) -> Script:`

KANAjetzt, line 724: Can you add a comment with a description what this function does?

addons/mod_loader/mod_loader.gd
```diff
@@ -703,12 +705,44 @@ func _apply_extension(extension_path)->Script:
...
+	if not File.new().file_exists(parent_script_path):
+		ModLoaderUtils.log_error("The parent script path '%s' does not exist" % [parent_script_path], LOG_NAME)
+		return null
+
+	# Check if the script to reset has been extended
+	if not _saved_scripts.has(parent_script_path):
+		ModLoaderUtils.log_error("The parent script path '%s' has not been extended" % [parent_script_path], LOG_NAME)
+		return null
+
+	# Check if the script to reset has been extended
```
KANAjetzt, line 735: I think that comment should be different? Based on the error message.

KANAjetzt (commented):
Looks good - ty 👍 

Only the question who will use the `_reset_extension()`? 
Mod authors?

otDan:
> Looks good - ty 👍
> 
> Only the question who will use `_reset_extension()`? Mod authors?

the use case will be mostly internal, it's just the base for the next prs, but it can be used in very specific use cases if a mod author wants to reset something specific, I suggest keeping it as internal

KANAjetzt (approved):
👍

Qubus0 (approved):
looks good.
`_saved_scripts{}` is essentially the same as `script_extensions[]`, right? The format of it is quite a bit different, so I'm not sure if it's possible or worth it to join the two. Just something to keep in mind

otDan:
> looks good. `_saved_scripts{}` is essentially the same as `script_extensions[]`, right? The format of it is quite a bit different, so I'm not sure if it's possible or worth it to join the two. Just something to keep in mind

well I may look into merging the two in later prs, but from what I have seen, `script_extensions[]` is used for ordering in the setup, might be worth a look into though

Issue #166, opened 2023-03-03
Discussion: String Standards


Qubus0:
My feelings:
prefer double quotes over single quotes: `"string"`, not `'string'`
prefer quote escaping over single quotes : `"\"hello\" world"`, not `'"hello" world'` (see comment below)
~~prefer single quotes around strings over quote escaping: `'"hello" world'`, not `"\"hello\" world"`~~

prefer format strings over string concatenation and `str()`: `"hello %s!" % place`, not `"hello " + place + "!"`, not `str("hello", place)`
except for very simple cases/single concatenation: `"hello " + place`, not `"hello %s" % place`

split long strings into shorter ones with string concatenation `"" + "" + ...`, not `str("", "", ...)`
```gdscript
print(
   "pretend that this" +
   "is a really long" +
   "message"
)
```

ithinkandicode:
I agree with everything here. 

Perhaps it would be good to create a new wiki page called Contributing, where we outline the standards we'd like everyone to follow.

Qubus0:
Typically that is done in CONTRIBUTING.md

KANAjetzt:
closed by https://github.com/GodotModding/godot-mod-loader/pull/229

PR #197, opened 2023-03-29
refactor: :recycle: add mod_id in all validation error messages

### LINKED ISSUE #173: ModManifest Validation - Make sure error messages include a reference to the mod that cause the validation error
As @ithinkandicode mentioned on Discord, the deprecation warning of `compatible_mod_loader_version` doesn't include the mod id, it is caused by. This requires the user to follow the stack trace and check the data himself. We can make that process easier by adding the mod id to the error message.

image

From a quick first look this functions need this treatment:

- [ ] `_handle_compatible_mod_loader_version()`
- [ ] `is_semver_version_array_valid()`
- [ ] `is_semver_valid()`

KANAjetzt:
- Added the mod id of the causing mod in all validation error messages
- Changed all single quotes to double quotes as discussed in #166
- Fixed a critical bug that caused the validation to break on `if is_semver_valid(version_number):`
Added the `not` to fix that.

closes #173

### DIFF
--- a/addons/mod_loader/classes/mod_manifest.gd
+++ b/addons/mod_loader/classes/mod_manifest.gd
@@ -77,7 +77,9 @@ func _init(manifest: Dictionary) -> void:
 	):
 		return
 
-	if is_semver_valid(version_number):
+	var mod_id = get_mod_id()
+
+	if not is_semver_valid(mod_id, version_number, "version_number"):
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
good catch on that error and good argument for unit tests..

PR #199, opened 2023-03-30
refactor: :recycle: changed mod config dir path to `user://configs`

### LINKED ISSUE #167: Move Config JSON files to user folder
Should configs save to `user://configs`, instead of `res://configs`?

I think saving to the user folder is probably better, because:

- It keeps the game directory clean. 
- It makes it less obvious where the data is held, which should make it less likely that a player will try to modify that data.

KANAjetzt:
- Adds `const MOD_CONFIG_DIR_PATH := "user://configs"` to *mod_loader_utils.gd*
- Updates `get_path_to_configs()` to use `MOD_CONFIG_DIR_PATH`

breaking: Any custom configs stored in `res://configs` will no longer be loaded. 

closes #167

### DIFF
--- a/addons/mod_loader/mod_loader_utils.gd
+++ b/addons/mod_loader/mod_loader_utils.gd
@@ -1,8 +1,10 @@
 class_name ModLoaderUtils
 extends Node
 
+
 const LOG_NAME := "ModLoader:ModLoaderUtils"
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
✨

PR #202, opened 2023-03-30
Fix mod id validation

Qubus0:
really simple check for length 3
@ithinkandicode do we really need all of that `type` stuff at the beginning of the method? what was the intended use for that and will that be followed up upon?

### DIFF
--- a/addons/mod_loader/classes/mod_manifest.gd
+++ b/addons/mod_loader/classes/mod_manifest.gd
@@ -370,7 +370,7 @@ static func is_mod_id_valid(original_mod_id: String, check_mod_id: String, type 
 	var check_namespace = split[0]
 	var check_name = split[1]
 	var re := RegEx.new()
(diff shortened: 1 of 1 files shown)


KANAjetzt:
What type stuff exactly? 
That one?

https://github.com/GodotModding/godot-mod-loader/blob/bc09c110c4bde727886e31a5cbbf2b9b852294a1/addons/mod_loader/classes/mod_manifest.gd#L354

KANAjetzt:
It's used in the error message to pass the field name.

https://github.com/GodotModding/godot-mod-loader/blob/bc09c110c4bde727886e31a5cbbf2b9b852294a1/addons/mod_loader/classes/mod_manifest.gd#L101-L104

Qubus0:
oh I missed the usage in that. disregard that comment then

KANAjetzt (approved):
Approved ✅ 
Do we consider this a breaking change? 😬

PR #198, opened 2023-03-29
feat: added _remove_extension

otDan:
Requires: https://github.com/GodotModding/godot-mod-loader/pull/195
Adds `_remove_extension` that allows for removal of a specific extension, this resets the base script and then reinstalls all the other extensions in the already sorted order

PR #203, opened 2023-03-31
feat: mod reloading

otDan:
This enables people to reload mods at runtime, currently the only changes made by mods that will be reset by this are extensions, which is the major changing factor in most cases. Opens up for next PRs that can reset nodes and hopefully all changes that mod developers want to make. 
An example use case scenario would be: user downloads or updates mod, without needing to restart the whole game, the user could click a button in game that reloads all mods. Another example could be a mod that introduces in game mod downloading and updating, allowing for a better user experience.
This PR also renames 
`_remove_extension` > `_remove_specific_extension_from_script` and 
`_reset_extension` > `_remove_all_extensions_from_script` for easier functionality comprehension.
Adds a fix for #198 that returns null in a return void function.

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -191,6 +191,21 @@ func _load_mods() -> void:
 	ModLoaderUtils.log_success("DONE: Installed all script extensions", LOG_NAME)
 
 
+# Internal call to reload mods
+func _reload_mods() -> void:
+	_reset_mods()
+	_load_mods()
+
+
+# Internal call that handles the resetting of all mod related data
+func _reset_mods() -> void:
+	mod_data.clear()
+	mod_load_order.clear()
+	mod_missing_dependencies.clear()
+	script_extensions.clear()
+	_clear_extensions()
+
+
 # Check autoload positions:
 # Ensure 1st autoload is `ModLoaderStore`, and 2nd is `ModLoader`.
 func _check_autoload_positions() -> void:
@@ -830,11 +845,25 @@ func install_script_extension(child_script_path:String):
 		_apply_extension(child_script_path)
 
 
+# Remove a specific script from the vanilla extension chain
+# This will remove only the specifically provided extension
+# and keep all other extensions of that vanilla script running
 func uninstall_script_extension(extension_script_path: String) -> void:
 
 	# Currently this is the only thing we do, but it is better to expose
 	# this function like this for further changes
 	_remove_specific_extension(extension_script_path)
+
+
+# This function should be called only when actually necessary
+# as it can break the game and require a restart for mods
+# that do not fully use the mod loader systems
+# Used to reload already present mods and load new ones
+func reload_mods() -> void:
+
+	# Currently this is the only thing we do, but it is better to expose
+	# this function like this for further changes
+	_reload_mods()
 
 
 # Register an array of classes to the global scope, since Godot only does that in the editor.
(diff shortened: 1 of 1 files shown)


addons/mod_loader/mod_loader.gd
```diff
@@ -796,6 +815,12 @@ func _reset_extension(parent_script_path: String) -> void:
 	_saved_scripts.erase(parent_script_path)
 
 
+func _clear_extensions() -> void:
+	var _to_remove_scripts = _saved_scripts.duplicate()
```
KANAjetzt, line 819: type

addons/mod_loader/mod_loader.gd
```diff
@@ -819,11 +844,25 @@ func install_script_extension(child_script_path:String):
...
 
 	# Currently this is the only thing we do, but it is better to expose
 	# this function like this for further changes
-	_remove_extension(extension_script_path)
+	_remove_specific_extension(extension_script_path)
+
+
+# This function should be called only when actually necessary
+# as it can break the game and require a restart for mods
+# that do not fully use the mod loader systems
```
KANAjetzt, line 859: what do you mean by `not fully use the mod loader systems`?

KANAjetzt (approved):
2 little things

addons/mod_loader/mod_loader.gd
KANAjetzt: `save_scene()` can be used to modify a scene and add nodes
https://github.com/GodotModding/godot-mod-loader/blob/80fea7219a2ad8cc187f4b7c2c0dbc8cbf1c03e7/addons/mod_loader/mod_loader.gd#L870-L876

And there is `append_node_in_scene()`:
https://github.com/GodotModding/godot-mod-loader/blob/80fea7219a2ad8cc187f4b7c2c0dbc8cbf1c03e7/addons/mod_loader/mod_loader.gd#L851C6-L867

Definitely room for improved here though.

otDan: yeah I mean mostly about non permanent node appending but yeah definitely have room for improvement and probably want to push mod developers to use them

KANAjetzt: Can you clarify what type of mods will break when reloading in the comment?
I guess every mod that consists of more then script extensions?

ithinkandicode:
I'm a bit behind on the latest development, but is my understanding of these funcs correct?

| Func | Purpose |
| --- | --- |
| `_remove_specific_extension` | Removes an extended script from the given script |
| `_remove_all_extensions` | Removes all extended scripts from the given script  |
| `_clear_extensions` | Resets **every** script, removing all extensions from every script |

If that's right, I'm wondering if the first 2 funcs could be affixed with `_from_script`? Because the func name `remove_all_extensions` in particular sounds very similar to `clear_extensions`, as in, one might presume that "remove all extensions" means removing all extensions from all scripts. That was my initial thought when I saw the func name. Perhaps the `_clear_extensions` func name could be clearer too, eg. `_remove_all_extensions_from_all_scripts`?

So my proposed func names would be this:

| Func | Purpose |
| --- | --- |
| `_remove_specific_extension_from_script` | Removes *a single* extended script from the given script |
| `_remove_all_extensions_from_script` | Removes *all* extended scripts from the given script  |
| `_remove_all_extensions_from_all_scripts` | Removes *all* extended scripts from *every* script  |

KANAjetzt (approved):
Tested in brotato no errors 👍
I did not test the `reload_mods()` func

Issue #161, opened 2023-03-02
ModLoaderStore - Singleton for to store data


ithinkandicode:
As @KANAjetzt suggested on Discord, if we add another singleton before `ModLoader`, we can use it to hold all the data that we might want to share between ModLoader core and all its associated methods.

This fixes an issue where you can't refer to the local variables in *mod_loader.gd* during its `_init` phase, because it's not ready yet.

This will help us to address these issues, which are currently (mostly) blocked by this problem:

- #109
- #140
- #141
- #146
- #145 (implement the remaining options)
- #153

It may also provide us with more optimisation options, eg for `_get_verbosity` (_mod_loader_utils.gd_) we can store the command line arg overrides once and refer back to them later, rather than checking them every time a log is performed.

Additionally, we can refactor out code to be simpler with less repetition, eg. currently in the 2 places that `get_local_folder_dir("configs")` is used, it's immediately followed by a check to see if there's a CLI override; we could simply store the path to use once, then refer to that single point of data instead.

ithinkandicode:
- Closed by #172

Issue #153, opened 2023-02-28
Splitting up long files (utils/core)


ithinkandicode:
| Done | Dir | File | Class Name | Notes |
| --- | --- | --- | --- | ---
| ✅ | `api` | **log.gd** | ModLoaderLog | done in #205 |
|✅| `api` | **mod.gd** | ModLoaderMod | All helper functions used by mod devs - #208 `ModLoader` ->  `ModLoaderMod` |
| ✅ | `api` | **config.gd** | ModLoaderConfig | done in #157 |
| ✅ | `api` | **deprecated.gd** | ModLoaderDeprecated |done in #156 |
| ✅ | `internal` | **cli.gd** | _ModLoaderCLI | #226 |
| ✅ | `internal` | **path.gd** | _ModLoaderPath | #226 |
| ✅ | `internal` | **file.gd** | _ModLoaderFile | Including the saving funcs from [#143](https://github.com/GodotModding/godot-mod-loader/pull/143). Maybe JSON funcs too? - done in #226 |
| ✅ | `internal` | **dependencies.gd** | _ModLoaderDependencies | done in #222  |
| ✅  | `internal` | **script_extension.gd** | _ModLoaderScriptExtension | done in #225 |
| ✅ | `internal` | **godot.gd** | _ModLoaderGodot | Eg. `check_autoload_position()` - done in #172 |
| ✅ |  `internal/third_party` |**steam.gd** | _ModLoaderSteam | See related chat in [#148](https://github.com/GodotModding/godot-mod-loader/pull/148), [here](https://github.com/GodotModding/godot-mod-loader/pull/148#discussion_r1119485922) - done in #158 |

### Breaking Changes & Deprecation

Ofc, if we moved the logger funcs then every currently published Brotato mod would break. But we could handle this with a deprecation notice, which can be handled with a new func and that triggers a fatal error in the editor. I've explained how this could work in more detail in #154:

- #154

ithinkandicode:
Additionally, perhaps it would be a good idea to also split up mod_loader.gd? Especially the public API methods.

Here's a table with my notes on class names and grouping (all applies to _mod_loader.gd_ except `ModLoaderLog`):

| Class | Notes |
| --- | --- |
| `ModLoaderCore` | Private methods |
| `ModLoaderAPI` | Public methods |
| `ModLoaderConfig` | Config JSON |
| `ModLoaderOptions` | ML Options (PR #145) |
[...]

PR #205, opened 2023-03-31
refactor: :recycle: Logging

### LINKED ISSUE #140: Logging: Track logged messages
In #138 I added the array variable `logged_messages`. I think this is a potentially very useful approach that could be applied it to all logged messages.

**Implementation:** 

- In `_loader_log`, we save logged messages to an array.
    - This is the same as what I did for #138, 
    - but with the array variable being in `ModLoaderUtils` instead. 
- For each message, we can track all the data parts that are passed as variables to `_loader_log`
  - Ie. mod name, message, and type.
  - Actually, it might be better as a dictionary, with these keys:
      - `all` = All logged messages
      - `by_mod` = Notices logged via a certain mod
      - `by_type` = Notices of a specific type

**API Methods:**

- `get_tracked_messages_all()`
- `get_tracked_messages_by_mod(mod_name: String)`
- `get_tracked_messages_by_type(type: String)` -- where `type` is one of the current log types (`fatal-error`, `error`, etc)

**Use Case:**

This would be a massive help to players who are using mods, as it would save them needing to check the logs (and save them needing to know how to do that). This does depend on the game or a mod presenting this data, but this is the first step towards supporting that.

Note: This isn't related to UI at all. It simply provides data that a UI could interact with, which would be the responsibility of the game developer or modders -- like how Brotato's Mods screen interacts with ModLoader's current data (see example with otDan's mod [BetterModList](https://github.com/otDan/Brotato-BetterModList)). So on that screen, it could include a console-like tray with any errors or warnings.

KANAjetzt:
# Logging refactor

- Moved log utils into its one file (*res://addons/mod_loader/api/log.gd*) and Class (`ModLoaderLog`).
- New `Resource` for log entries `ModLoaderLogEntry`
- New util functions to work with stored logs
   - `get_all_as_resource_array()`
   - `get_all_as_string_array()`
   -  `get_by_mod_as_resource_array(mod_name: String)`
   -  `get_by_mod_as_string_array(mod_name: String)`
   - `get_by_type_as_resource_array(type: String)`
   -  `get_by_type_as_string_array(type: String)`

## Todos 
- [x] **Move functions to `ModLoaderLog`**
   - [x] `_write_to_log_file`
   - [x] `_rotate_log_file`
   - [x] `_clear_old_log_backups`   
   - [x] `_is_mod_name_ignored`
   - [x] `_get_verbosity`
   - [x] `_loader_log`
   - [x] `_get_time_string`
   - [x] `_get_date_string`
   - [x] `_get_date_time_string`
- [x] **Move functions to `ModLoaderLog` and leave `deprecation()` func**
   - [x] `log_fatal` -> `ModLoaderLog.fatal`
   - [x] `log_error` -> `ModLoaderLog.error`
   - [x] `log_warning` -> `ModLoaderLog.warning`
   - [x] `log_info` -> `ModLoaderLog.info`
   - [x] `log_success` -> `ModLoaderLog.success`
   - [x] `log_debug` -> `ModLoaderLog.debug`
   - [x] `log_debug_json_print`  -> `ModLoaderLog.debug_json_print`
- [x] Add `ignored_mod_names_in_log` to `ml_options` in `ModLoaderStore` and *options_profile.gd*
- [x] update `get_mod_config()` in *config.gd* line 73 and 75
- [x] Add `get_all_as_resource()`
- [x] Add `get_all_as_string()`
- [x] Add `get_by_mod_as_resource(mod_name: String)`
- [x] Add `get_by_mod_as_string(mod_name: String)`
- [x] Add `get_by_type_as_resource(type: String)`
- [x] Add `get_by_type_as_string(type: String)`
- [x] **Error Handling**
   - [x] Check if key exists in `get_by_mod(mod_name: String)`

## Test Mod
[KANA-LogViewer.zip](https://github.com/GodotModding/godot-mod-loader/files/11185148/KANA-LogViewer.zip)

<br />
<br />

*closes #140*


KANAjetzt:
Oh and I still need to fix this:
- https://github.com/GodotModding/godot-mod-loader/issues/140#issuecomment-1490453112

> Working on this now. Just so I don't forget update `get_mod_config()` in _config.gd_ line 73 and 75:
> 
> ```gdscript
> if not ModLoaderStore.logged_messages.has(full_msg):
>    ModLoaderUtils.log_debug(full_msg, mod_dir_name)
>    ModLoaderStore.logged_messages.push_back(full_msg)
> ```

ithinkandicode:
I was thinking the log funcs would be moved to `./add_ons/mod_loader/api/log.gd`? And the class would follow the standard of other API classes, ie. `ModLoaderLog`. That follows the established standard and retains the namespaces classes.

Note: The current `classes` directory might be misnamed (my fault), as it's actually only used for **resources**. The functional classes are actually in `api` now.

ithinkandicode:
I've looked over this PR's code in more detail and I have some notes.

There are still logging methods in `ModLoaderUtils`. Can they all be moved to `ModLoaderLog` please? And leave deprecation funcs behind -- eg. see [#208](https://github.com/GodotModding/godot-mod-loader/pull/208), which moves methods from `ModLoader.*` to `ModLoaderMod.*`

How come you've used a class instance for LogManager? Why not use it with static methods, like other API methods do? Then you can log things with `ModLoaderLog.method`, rather than `ModLoaderStore.LogManager.method`.

Generally I think we should keep the `ModLoaderFoo` standard of class names. This makes it easier to identify methods that originate with ModLoader, and avoids potential conflicts with games/mods that use more generic class names (eg. an existing game might already have a class called _LogManager_, but it definitely won't have a class called _ModLoaderLog_).

You've used ModLoaderStore to store a class instance, but it's not intended to be used that way. It's meant to be a simple data store, and was created to fix these two issues:

1. You couldn't access the local variables from ModLoader.* during ModLoader's init phase. See issue [#161](https://github.com/GodotModding/godot-mod-loader/issues/161) and PR [#172](https://github.com/GodotModding/godot-mod-loader/pull/172) for more details on this. This point isn't related to the PR but I'm mentioning it here for completeness.
1. Static methods can't save data to their own class (because that class hasn't been instanced). So now instead, they can store data to ModLoaderStore. In this case, ModLoaderStore lets you keep track of any logged messages, via `ModLoaderStore.logged_messages`. So all your static methods can still exist in ModLoaderLog, but you can also save data, like logged messaged, to ModLoaderStore.

KANAjetzt:
I will rework this to account for #153.
Ty @ithinkandicode for the review 👍

ithinkandicode:
Should this PR be called "Logging Refactor", instead of "store logged messages"? Just thinking of the commit message which is set automatically from the PR title

KANAjetzt:
good point

addons/mod_loader/setup/setup_log.gd
```diff
@@ -0,0 +1,214 @@
+class_name ModLoaderSetupLog
```
ithinkandicode, line 1: I like the idea of splitting up the setup file, but wouldn't it better in its own PR? Ideally I think we want PRs to have a singular focus, eg "move log funcs" without them affecting things beyond the scope stated in the PR's title. This way it makes them easier to search for on GitHub, and makes the PR's commit message clear when looking through the Git history.

addons/mod_loader/setup/setup_utils.gd
```diff
@@ -0,0 +1,190 @@
+class_name ModLoaderSetupUtils
```
ithinkandicode, line 1: A lot of this code seems to be duped when I don't know if it needs to be. Is there not a better way to handle this without duplicating code?

addons/mod_loader/setup/setup_log.gd
KANAjetzt: Yes I feel like this PR got a bit out of hand again. 
I look into how I can split it up.

addons/mod_loader/setup/setup_utils.gd
KANAjetzt: I tried my best.

> The trade-offs for using the existing ModLoader Classes in the setup script are to big. So the required functions where moved into the *setup/* dir.

The current issue is:

- `ModLoaderUtils` is now depending on other classes that don't get loaded by default when using it in the setup script
- The issue with `ModLoaderStore` not being available during setup and no way to check if the class exists ( `Engine.get_singleton()` is apparently only working for build in Autoloads )

addons/mod_loader/setup/setup_log.gd
KANAjetzt: done *[cfd35d4](https://github.com/GodotModding/godot-mod-loader/pull/205/commits/cfd35d42dfeb90d663a089231ef4bd95f063206d)*

addons/mod_loader/api/log.gd
```diff
@@ -0,0 +1,385 @@
...
+	func get_entry() -> String:
+		return time + get_prefix() + message
+
+
+	func get_prefix() -> String:
+		return "%s %s: " % [type.to_upper(), mod_name]
+
+
+	func get_md5() -> String:
+		return str(get_prefix(), message).md5_text()
```
Qubus0, line 45: seems a bit inconsistent to use str() here and + a few lines above

addons/mod_loader/api/log.gd
```diff
@@ -0,0 +1,385 @@
...
+	if not error == OK:
+		assert(false, "Could not open log file, error code: %s" % error)
+		return
+
+	log_file.seek_end()
+	log_file.store_string("\n" + string_to_write)
+	log_file.close()
+
+
+# Keeps log backups for every run, just like the Godot; gdscript implementation of
```
Qubus0, line 318: `like the Godot;` ?

addons/mod_loader/mod_loader_store.gd
```diff
@@ -59,6 +70,9 @@ var ml_options := {
 	# error. This can be helpful when developing mods that depend on a mod that
 	# hasn't been updated to fix the deprecated issues yet
 	ignore_deprecated_errors = false,
+
+	# TODO: Description goes here
```
Qubus0, line 74: ☝️

Qubus0 (approved):
only minor things. 
very nice work 🚀

addons/mod_loader/api/log.gd
KANAjetzt: good question I just copy pasted the original 😄

*fixed https://github.com/GodotModding/godot-mod-loader/pull/205/commits/3bff9e9bd215d00c716cac70eed152e38df29966*

addons/mod_loader/mod_loader_store.gd
KANAjetzt: *fixed https://github.com/GodotModding/godot-mod-loader/pull/205/commits/3bff9e9bd215d00c716cac70eed152e38df29966*

addons/mod_loader/api/log.gd
KANAjetzt: I did it so I can call `.md5_text()` right away.
Changed `get_entry()` to use `str()` as well.

*fixed https://github.com/GodotModding/godot-mod-loader/pull/205/commits/0c81c88ad7655dd684c139be656ed769bf620324*

PR #209, opened 2023-04-09
Add PR Dependency Check GitHub Action

KANAjetzt:
Add PR Dependency Check GitHub Action

https://github.com/marketplace/actions/pr-dependency-check


Qubus0 (approved):
good idea

Issue #141, opened 2023-02-26
Logging: Add arg for `only_once`


ithinkandicode:
Add an arg to all the log methods that let you only log the message once. 

This would provide a standardised way to implement what I did for #138.

Doing this may depend on implementing #140 first.

**Related:**

- #138 
- #140

ithinkandicode:
- Closed by #210

PR #210, opened 2023-04-09
feat: ✨ Added `only_once` parameter to all log funcs

KANAjetzt:
Depends on #205

Adds an arg to all the log methods that let you only log the message once.

closes https://github.com/GodotModding/godot-mod-loader/issues/141


Qubus0 (approved):
beautiful ✨

PR #211, opened 2023-04-09
refactor: :recycle: created *setup_log.gd* and setup_utils.gd*

KANAjetzt:
The tradeoffs of utilizing the existing ModLoader Classes in the setup script are too significant. So the required functions where moved into the *setup/* dir and any dependencies to other classes where removed.


Qubus0 (approved):
I wish it could be more beautiful, but I guess we'll have to live with this necessary evil for now.
good work nonetheless

PR #212, opened 2023-04-10
Delete .github directory

KANAjetzt:
We actually only need this on the dev branch 👀


Qubus0 (changes requested):
if we merge dev into main it will be back. 
also not this from the docs
> A workflow run is triggered for any workflows that have on: values that match the triggering event. Some events also require the workflow file to be present on the default branch of the repository in order to run.
https://docs.github.com/en/actions/using-workflows/triggering-a-workflow

filtering out main is better
```yaml
on:
  pull_request:
    branches-ignore:
      - main
```
https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-excluding-branches

KANAjetzt:
Right for some reason I thought the `.github/` dir is ignored on merge.

PR #214, opened 2023-04-12
feat: ✨ Add user profiles

### LINKED ISSUE #109: Add option to disable certain mods via JSON
We could add support for checking a file called `mods.json` in the user data folder, and when ModLoader inits, we could check that file and skip any disabled mods.

We could also provide a func to save the settings to this file.

This would let games implement a toggle to enable/disable certain mods from within the game itself, so users don't need to unsubscribe from them in the workshop/Thunderstore, or manually delete/rename mod ZIPs from their game folder.

KANAjetzt:
# Add User Profiles - `ModLoaderUserProfile`

## Inner Class `Profile`
Stores the data for each User Profile
```gdscript
	var name := ""
	var mod_list := {}
```

## New API Methods

- `enable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile)`
Enables a mod - it will be loaded on the next game start

- `disable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile)`
Disables a mod - it will not be loaded on the next game start

- `delete_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile)`
Delete a mod from a user profiles mod_list

- `create(profile_name: String)`
 Creates a new user profile with the given name, using the currently loaded mods as the mod list.

- `set_profile(profile_name: String)`
Sets the current user profile to the specified profile_name.

- `delete(profile_name: String)`
Deletes a user profile with the given profile_name.

- `get_current()`
Returns the current user profile

- `get_profile(profile_name: String)`
Return the user profile with the given name

- `get_all_as_array()`
Returns an array containing all user profiles stored in ModLoaderStore

## New Internal Methods

- `_update_disabled_mods()`
Update the global list of disabled mods based on the current user profile. The user profile will override the `disabled_mods` property that can be set via the options resource in the editor. 
Example: If `Mod-TestMod` is set in `disabled_mods` via the editor, the mod will appear disabled in the user profile. If the user then enables the mod in the profile the entry in `disabled_mods` will be removed.
Used in *mod_loader.gd* `_init()` before loading mods.

- `_update_mod_lists()`
This function updates the mod lists of all user profiles with newly loaded mods that are not already present. It does so by comparing the current set of loaded mods with the mod list of each user profile, and adding any missing mods.
Used in *mod_loader.gd* `_ready()` after all mods are loaded.

- `_handle_mod_state(mod_id: String, profile_name: String, activate: bool)`
Handles the activation or deactivation of a mod in a user profile.

- `_is_mod_id_in_mod_list(mod_id: String, profile_name: String)`
Checks whether a given mod_id is present in the mod_list of the specified user profile.
Returns True if the mod_id is present, False otherwise.

- `_create_new_profile(profile_name: String, mod_list: Dictionary)`
Creates a new Profile with the given name and mod list.
Returns the newly created Profile object.

- `_load()`
Loads user profiles from the JSON file and adds them to ModLoaderStore.
Path to JSON file stored in `const FILE_PATH_USER_PROFILES`

-  `_save()`
Saves the user profiles in the ModLoaderStore to the user profiles JSON file.
Path to JSON file stored in `const FILE_PATH_USER_PROFILES`

## Locked mods
*Renamed from `mandatory_mods`*

- Added a new property to `ModData` - `is_locked`
The `is_locked` flag indicates whether a mod's `active` state can be changed in a user profile, effectively preventing the user from disabling or enabling the mod.
- Added `locked_mods` to *options_profile.gd*
- Added `locked_mods` to `ModLoaderStore.ml_options`

## Test / Example Mod: 
image
### Download [KANA-UserProfile.zip](https://github.com/GodotModding/godot-mod-loader/files/11295869/KANA-UserProfile.zip)

<br/>
<br/>

--- 

*closes #109*


addons/mod_loader/api/profile.gd
```diff
@@ -0,0 +1,335 @@
...
+# Internal profile functions
+# =============================================================================
+
+# Update the global list of disabled mods based on the current user profile
+# The user profile will override the disabled_mods property that can be set via the options resource in the editor.
+# Example: If "Mod-TestMod" is set in disabled_mods via the editor, the mod will appear disabled in the user profile.
+# If the user then enables the mod in the profile the entry in disabled_mods will be removed.
+static func _update_disabled_mods() -> void:
+	var user_profile_disabled_mods := []
+	var current_user_profile : Profile
```
KANAjetzt, line 176: `var current_user_profile : Profile` -> `var current_user_profile: Profile`

addons/mod_loader/api/profile.gd
```diff
@@ -0,0 +1,335 @@
...
+	return _handle_mod_state(mod_id, profile_name, true)
+
+
+# Disables a mod - it will not be loaded on the next game start
+static func disable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile) -> bool:
+	# Check if it is a mandatory mod
+	if ModLoader.mod_data.has(mod_id) and ModLoader.mod_data[mod_id].is_mandatory:
+		ModLoaderLog.error(
+			"Unable to disable mod \"%s\" as it is marked as mandatory \"%s\" and cannot be deactivated."
+			% [mod_id, ModLoaderStore.ml_options.mandatory_mods],
```
Qubus0, line 31: a little confusing wording. and `mandatory_mods` is an array, right? maybe something like
```
"Unable to disable mod \"%s\" as it is marked as mandatory. Mandatory mods: %s"
```
Also what exactly does mandatory mean? that another mod depends on it? then it might be a good idea to mention the mod that depends on it in case someone wants to toggle it off.
I am assuming that the activation/deactivation interface would listen to the logs from here for errors and display them to the user

addons/mod_loader/api/profile.gd
```diff
@@ -0,0 +1,335 @@
...
+		LOG_NAME)
+		return false
+
+	return _handle_mod_state(mod_id, profile_name, false)
+
+
+# Delete a mod from a user profiles mod_list
+static func delete_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile) -> bool:
```
Qubus0, line 39: What exactly is the difference between a disabled mod and a deleted mod?
is a deleted mod just disabled and will not show up in the list? Or is this just to remove any mod that isn't currently installed? if that's the case, it might make sense to make this private, so a UI will only enable and disable mods

addons/mod_loader/api/profile.gd
```diff
@@ -0,0 +1,335 @@
...
+	var is_save_success := _save()
+
+	if is_save_success:
+		ModLoaderLog.debug("Current user profile set to \"%s\"" % profile_name, LOG_NAME)
+
+	return is_save_success
+
+
+# Deletes a user profile with the given profile_name.
+static func delete(profile_name: String) -> bool:
```
Qubus0, line 114: with both mods and profiles having methods, it makes sense to explicitly name them `delete_mod` and `delete_profile` (also for other methods)

addons/mod_loader/api/profile.gd
```diff
@@ -0,0 +1,335 @@
...
+	var is_save_success := _save()
+
+	if is_save_success:
+		ModLoaderLog.debug("Updated the mod lists of all user profiles", LOG_NAME)
+
+	return is_save_success
+
+
+# Handles the activation or deactivation of a mod in a user profile.
+static func _handle_mod_state(mod_id: String, profile_name: String, activate: bool) -> bool:
```
Qubus0, line 227: "handle" always sounds like it is reacting to the change of something. this feels more like a "set_mod_state", a method that changes something itself

addons/mod_loader/api/profile.gd
```diff
@@ -0,0 +1,335 @@
...
+static func _is_mod_id_in_mod_list(mod_id: String, profile_name: String) -> bool:
+	# Get the user profile
+	var user_profile := get_profile(profile_name)
+	if not user_profile:
+		# Return false if there is an error getting the user profile
+		return false
+
+	# Return false if the mod_id is not in the profile's mod_list
+	if not user_profile.mod_list.has(mod_id):
+		ModLoaderLog.error("Mod id \"%s\" not found in the \"mod_list\" of user profile \"%s\"." % [mod_id, profile_name], LOG_NAME)
```
Qubus0, line 255: i'm not sure if the error message belongs in the method that checks the state, unless it's never used for anything else

Qubus0 (commented):
very nice stuff, good work 👍 
i think the user profiles mod is such a central thing that we might want to pick it up as an organisation repo
also: This seems awesome for implementing mod packs in the future. just create a profile and provide an api to download all of those from the steam workshop

addons/mod_loader/api/profile.gd
KANAjetzt: I briefly mentioned it in the PR description:
> Mandatory mods
Maybe should be renamed to locked_mods? - So it can be used to not allow deactivation or activation of a mod.

I added it because it would be unfortunate if you deactivated the user profile UI mod😅.

addons/mod_loader/api/profile.gd
Qubus0: check if the files are there?

KANAjetzt: fair point 😄

addons/mod_loader/api/profile.gd
Qubus0: in that context it's like a mod that is currently in use? but will we use it for anything else as well?

addons/mod_loader/api/profile.gd
KANAjetzt: I think as long as it's only used in the local scope, it should be fine. 
It can be changed later if it turns out to be an issue.

addons/mod_loader/api/profile.gd
KANAjetzt: *fixed https://github.com/GodotModding/godot-mod-loader/pull/214/commits/0afe2b4d676b0c703ba4983e3062e95d67c1429d*

addons/mod_loader/api/profile.gd
KANAjetzt: *done - https://github.com/GodotModding/godot-mod-loader/pull/214/commits/bb8db97b135eb04d2a1f60e5ac40b0defcabf4f6*

addons/mod_loader/api/profile.gd
KANAjetzt: Changed it to `is_locked` and `locked_mods` - this allows a mod to be 'locked' in the current user profile state. So it can't be disabled or enabled by the user.

*https://github.com/GodotModding/godot-mod-loader/pull/214/commits/ebed6d8b8dd9f68563b7c215589b0976036d2255*

Qubus0 (approved):
nice work!

PR #215, opened 2023-04-14
fix: :bug: check if a mod `is_loadable`

KANAjetzt:
Mods should not pass the dependency check if one of there dependencies are not loadable.


Qubus0 (approved):
good catch

PR #216, opened 2023-04-16
test: :white_check_mark: tests for manifest validation

KANAjetzt:
Tests for:

- `is_mod_id_valid()` by @Qubus0 
- `validate_distinct_mod_ids_in_arrays()`
- `test_is_mod_id_array_valid()`

### DIFF
--- a/test/Unit/test_manifest_validation.gd
+++ b/test/Unit/test_manifest_validation.gd
@@ -1,0 +1,74 @@
+extends GutTest
+
+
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
looks good for now, though i'd like to use the `params = use_parameters(...)` way, since that also creates a test case that can fail individually.
What I wold like to see is the `.gitattributes` file, so downloading via zip/asset lib won't contain all the tests

PR #217, opened 2023-04-17
chore: Add `.gitattributes` to allow AssetLib downloads

KANAjetzt:
To provide the mod loader on the Godot AssetLib we need to get rid of any third-party dependencies. Fortunately, we only need these tools for the self setup.


Qubus0 (approved):
nice 👍

PR #223, opened 2023-04-19
chore: :truck: Copied LICENSE into *addons\mod_loader*

KANAjetzt:
I copied the license file into the *addons/mod_loader* directory so that it is in the correct location for AssetLib downloads. However, I also left a copy in the root of the repository so that Github can still recognize it.

Added the LICENSE in the repo root to `.gitattributes`


Qubus0 (approved):
good

Issue #227, opened 2023-04-20
Naming Convention


KANAjetzt:
- Prefix local (private) to file vars / funcs with `_`
- Prefix classes that should only be used by the ModLoader with `_`
- If a method is in a non-prefixed class and ModLoader Internal, but used outside of the private scope, prefix with `_`, use it outside the scope, but add a comment why it was used there

Qubus0:
Reasoning: 
1. underscore methods/vars should only be used within the same file
2. most classes should not be used by mods, only by the modloader. if they are prefixed with an underscore, no mod should access them and we are free to change the internal structure without breaking mods and needing deprecations
3. in some cases we need to use private methods outside of their file (`_rotate_log_file` for example) and the class is a public one (`ModLoaderLog` here). Since the method should not be accessible to mods, we are using a "private" method outside of its scope here - and that needs an explanation

PR #232, opened 2023-04-21
style: ✏️ rename funcs to match naming convention

KANAjetzt:
Updated all files to match the naming convention we agreed upon in #227.


Qubus0 (approved):
✨

PR #233, opened 2023-04-21
refactor: ♻️ move consts from `ModLoader` to `ModLoaderStore`

KANAjetzt:
### Disclaimer
Not sure if we should do this. Moving `const UNPACKED_DIR` will break a lot of Brotato Mods because they use this pattern in there *mod_main.gd* most of the time:
`dir = modLoader.UNPACKED_DIR + "Darkly77-ContentLoader/"`

And I don't think there is a good way to add a deprecation warning to a `const`?

---

- Moved to `ModLoaderStore`:
   - `MODLOADER_VERSION`
   - `DEBUG_ENABLE_STORING_FILEPATHS`
   - `UNPACKED_DIR`
   - `REQUIRE_CMD_LINE`
- Added `get_unpacked_dir()` to `ModLoaderMod`
- Added `get_unpacked_mods_dir_path()` to `_ModLoaderPath`

<br/>

*works towards https://github.com/GodotModding/godot-mod-loader/issues/153*

### DIFF
--- a/addons/mod_loader/api/mod.gd
+++ b/addons/mod_loader/api/mod.gd
@@ -94,3 +94,7 @@ static func save_scene(modified_scene: Node, scene_path: String) -> void:
 	packed_scene.take_over_path(scene_path)
 	ModLoaderLog.debug("save_scene - taking over path - new path -> %s" % packed_scene.resource_path, LOG_NAME)
 	ModLoaderStore.saved_objects.append(packed_scene)
+
+
+static func get_unpacked_dir() -> String:
+	return _ModLoaderPath.get_unpacked_mods_dir_path()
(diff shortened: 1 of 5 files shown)


Qubus0:
found a way to deprecate a const via setget. it has to be made into a var though.
```gdscript
var TEST := "This is a test const" setget ,deprecated_direct_access_const_test

func deprecated_direct_access_const_test():
	ModLoaderDeprecated.deprecated_message("The const TEST was removed, use [proper method] instead")
	return "AAA" # proper method here
```

KANAjetzt:
ty @Qubus0 - for some reason it didn't cross my mind that I can just change the const to a var 😄 
I added the deprecation warning 👍

Qubus0 (approved):
👍

PR #252, opened 2023-05-26
refactor: ♻️ Ignore hidden directories

Ategon:
Currently the mod loader will try to read hidden directories (such as .git) as if they were mods and then throw assertion errors. This just adds a small check to stop that

### DIFF
--- a/addons/mod_loader/mod_loader.gd
+++ b/addons/mod_loader/mod_loader.gd
@@ -350,15 +350,14 @@ func _setup_mods() -> int:
 			# Stop loading mod zip files
 			break
 
-		# Only check directories
-		if not dir.current_is_dir():
-			continue
-
-		if mod_dir_name == "." or mod_dir_name == "..":
-			continue
-
-		# Ignore hidden directories
-		if mod_dir_name.begins_with("."):
+		if (
+			# Only check directories
+			not dir.current_is_dir()
+			# Ignore self and hidden directories
+			or mod_dir_name.begins_with(".")
+			# Ignore parent directory link
+			or mod_dir_name == ".."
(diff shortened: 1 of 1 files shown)


otDan (approved):

addons/mod_loader/mod_loader.gd
```diff
@@ -357,6 +357,10 @@ func _setup_mods() -> int:
 		if mod_dir_name == "." or mod_dir_name == "..":
 			continue
 
+		# Ignore hidden directories
+		if mod_dir_name.begins_with("."):
```
KANAjetzt, line 361: While we're here, we can probably put these three together.


```gdscript
if (
   not dir.current_is_dir() or
   mod_dir_name.begins_with(".") or
   mod_dir_name == ".."
):
   continue
```

KANAjetzt (commented):
ty and welcome 🥇

Ategon:
Ive pushed a change for combining the three statements as said above. Moved the comments to be alongside them and have the operators at the start of the lines for clarity + diff reasons

aesereht:
```gd
if (
   not dir.current_is_dir() or
   mod_dir_name.begins_with(".") or
   mod_dir_name == ".."
):
   continue
```
Logically it makes sense, but after combining conditions isn't this check redundant? `mod_dir_name == ".."` as it should be already checked by `mod_dir_name.begins_with(".")`.

KANAjetzt:
good point @aesereht

Ategon:
Pushed another change to remove that last check

KANAjetzt (approved):
Nice - ty 👍

PR #218, opened 2023-04-18
refactor: :recycle: Move `mod_data` `ModLoader` -> `ModLoaderStore`

KANAjetzt:
# Move `mod_data` from `ModLoader` to `ModLoaderStore`

This moves `mod_data` from *mod_loader.gd* to *mod_loader_store.gd*, enabling easier access from different classes.

## Breaking

- Two mods in my list, because they directly access `ModLoader.mod_data`.
`if ModLoader.mod_data.keys().has("damiBenro-Junker")`
- The default mod menu in Brotato for the same reason.
`for key in ModLoader.mod_data:`

Opened #219 and #221 to prevent this in the future. 

<br />
<br />

---
*works towards #153*

PR #258, opened 2023-06-05
fix: 🐛 added deprecation for `mod_data`

### LINKED ISSUE #239: deprecate mod_data
in #218, `mod_data` was moved from ModLoader to ModLoaderStore
in #233, we found out a way to deprecate variables. the same method should be applied here

~~Additionally, we should provide a proper way to get information about wether a mod is installed and active or disabled.
something along the lines of `is_mod_installed` / `is_mod_active` in either ModLoaderMod or ModLoader itself~~
removed in favour of #219

KANAjetzt:
- Adds `get` function `deprecated_direct_access_mod_data()` to deprecate the direct access to `ModLoader.mod_data` 
*( mod_data was moved to `ModLoaderStore` in #218 )*
- Renames `get_mod_data_from_mod_id(mod_id: String)` to `get_mod_data(mod_id: String)`

<br/>
<br/>

closes #239


Qubus0 (approved):
looking good

PR #260, opened 2023-06-05
refactor: :recycle: store Profile object instead of  profile name

### LINKED ISSUE #247: ModLoaderStore: store the Profile Object instead of the profile name in `current_user_profile`
https://github.com/GodotModding/godot-mod-loader/blob/97eda11fe144dd6c7ccc3b48db6b784e0eaf3b0a/addons/mod_loader/mod_loader_store.gd#L78-L79

KANAjetzt:
This will make it easier to access all profile data of the current profile. Before, you always had to get the profile data from `ModLoaderStore.user_profiles`.

- Removed the sub Class `Profile` in `ModLoaderUserProfile`
- Added new Resource Class `ModUserProfile`

To allow setting the type in `ModLoaderStore` without causing a cyclic dependency.

<br/>
<br/>

closes #247

### DIFF
--- a/addons/mod_loader/api/profile.gd
+++ b/addons/mod_loader/api/profile.gd
@@ -7,28 +7,22 @@ extends Object
 const LOG_NAME := "ModLoader:UserProfile"
 const FILE_PATH_USER_PROFILES = "user://mod_user_profiles.json"
 
-class Profile:
-	extends Reference
-
-	var name := ""
-	var mod_list := {}
-
 
 # API profile functions
 # =============================================================================
 
 # Enables a mod - it will be loaded on the next game start
-static func enable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile) -> bool:
+static func enable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile.name) -> bool:
 	return _set_mod_state(mod_id, profile_name, true)
 
 
 # Disables a mod - it will not be loaded on the next game start
-static func disable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile) -> bool:
+static func disable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile.name) -> bool:
 	return _set_mod_state(mod_id, profile_name, false)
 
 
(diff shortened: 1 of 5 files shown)


addons/mod_loader/api/profile.gd
```diff
@@ -79,15 +73,15 @@ static func create_profile(profile_name: String) -> bool:
...
+# Sets the current user profile to the profile with the specified profile_name.
 static func set_profile(profile_name: String) -> bool:
 	# Check if there is a user profile with the specified name
 	if not ModLoaderStore.user_profiles.has(profile_name):
 		ModLoaderLog.error("User profile with name \"%s\" not found." % profile_name, LOG_NAME)
 		return false
 
 	# Update the current_user_profile in the ModLoaderStore
-	ModLoaderStore.current_user_profile = profile_name
+	ModLoaderStore.current_user_profile = get_profile(profile_name)
```
Qubus0, line 84: this seems a little unintuitive now. normally a setter would take the whole object and set with that. here we are passing the name, looking up the whole object and then setting it. 
maybe in this case just naming it `set_profile_by_name` (or similar) could clear it up.

otDan (approved):
looks all good!

Qubus0 (approved):
makes a lot more sense now 👍

addons/mod_loader/api/profile.gd
KANAjetzt: changed it in *[c78897c](https://github.com/GodotModding/godot-mod-loader/pull/260/commits/c78897cad685013bee76ef1b317767db66ecc693)*

- `disable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile.name)`
- `enable_mod(mod_id: String, profile_name := ModLoaderStore.current_user_profile.name)`
- `set_mod_current_config(mod_id: String, config_name: String, profile_name := ModLoaderStore.current_user_profile.name)`
- `delete_profile(profile_name: String)`

All of them now accept the user profile resource as an argument instead of just the name. If the need arises, we can later add methods with `_by_name`.

PR #264, opened 2023-06-08
feat: ✨ option to allow autoload placement anywhere

### LINKED ISSUE #257: Allow the ModLoader autoload to be in a different spot than first
Issue: some games (Dome Keeper) implement cheat detection with autoloads. Placing the ModLoader after those will make modding them a lot harder.

* allow the modloader and store to be in a different spot than first
* still make sure that the store is added before the loader
* possibly add a log message that mentions that all autoloads above the mod loader won't be moddable if it is not in the first spot

this issue goes hand in hand with #250 , but is easier to implement for now.

### LINKED ISSUE #262: `check_autoload_position()` for `ModLoaderStore` not able to log error
Due to the significant dependency on `ModLoaderStore`, the current implementation of `_ModLoaderGodot.check_autoload_position()` is unable to log the intended error message.

```
`_ModLoaderGodot.check_autoload_position()`
  `ModLoaderLog.fatal()`
     `ModLoaderLog._log()` 
        -> `ModLoaderLog._is_mod_name_ignored()`
 ```
 
`ModLoaderLog._is_mod_name_ignored()` retrieves an array of ignored mod names from `ModLoaderStore`:
 
 ```gdscript
 static func _is_mod_name_ignored(mod_name: String) -> bool:
	var ignored_mod_names := ModLoaderStore.ml_options.ignored_mod_names_in_log as Array

	if not ignored_mod_names.size() == 0:
		if mod_name in ignored_mod_names:
			return true
	return false
 ```

KANAjetzt:
To make things work when `ModLoaderStore` is not placed before `ModLoader`, I had to move the check to the beginning of the `_init()` function. To handle the problem in `check_autoload_order()`, where I couldn't use `ModLoaderLog.fatal()` due to its reliance on a function that references `ModLoaderStore`, I directly loaded the ModLoader Options resource in `_check_autoload_positions()`. Check out issue https://github.com/GodotModding/godot-mod-loader/issues/262 for more info.

Did the same fix for `check_autoload_position()` to allow the error logging without `ModLoaderStore`

<br/>
<br/>

closes #257
closes #262

### DIFF
--- a/addons/mod_loader/classes/options_profile.gd
+++ b/addons/mod_loader/classes/options_profile.gd
@@ -9,6 +9,7 @@ export (bool) var enable_mods = true
 export (Array, String) var locked_mods = []
 export (ModLoaderLog.VERBOSITY_LEVEL) var log_level := ModLoaderLog.VERBOSITY_LEVEL.DEBUG
 export (Array, String) var disabled_mods = []
+export (bool) var allow_modloader_autoloads_anywhere = false
 export (bool) var steam_workshop_enabled = false
 export (String, DIR) var override_path_to_mods = ""
 export (String, DIR) var override_path_to_configs = ""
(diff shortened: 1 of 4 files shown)


Qubus0 (approved):
very nice, thank you

PR #265, opened 2023-06-08
chore: :pencil2: renamed `classes` dir to `resources`

### LINKED ISSUE #245: Rename "classes" directory to "resources"
The current name of "classes" is confusing, as nearly every file in ML is a class. But this directory only contains resources, so I suggest we change the name to that.

KANAjetzt:
Blocked by #260
Blocked by #261
Blocked by #264

> [ithinkandicode](https://github.com/ithinkandicode) commented [on May 7](https://github.com/GodotModding/godot-mod-loader/issues/245#issue-1699136300)
The current name of "classes" is confusing, as nearly every file in ML is a class. But this directory only contains resources, so I suggest we change the name to that.

closes #245


Qubus0 (approved):
✨

PR #269, opened 2023-06-11
fix: :bug: removed `dir_exists()` check

### LINKED ISSUE #267: `ModLoaderConfig` - Error when creating first default config
```
E 0:03:19.765   GDScriptFunctions::call: Encountered an error (12) when attempting to write to a file, with the path: 
  <C++ Source>  modules\gdscript\gdscript_functions.cpp:775 @ GDScriptFunctions::call()
  <Stack Trace> log.gd:215 @ _log()
                log.gd:62 @ fatal()
                file.gd:140 @ _save_string_to_file()
                file.gd:152 @ save_dictionary_to_json_file()
                mod_config.gd:72 @ save_to_file()
                mod_manifest.gd:228 @ load_mod_config_defaults()
                mod_data.gd:80 @ load_configs()
                mod_loader.gd:112 @ _load_mods()
                mod_loader.gd:67 @ _init()
```

`ERR_FILE_CANT_OPEN = 12 --- File: Can't open error.`

The issue is that `get_path_to_mod_configs_dir(mod_id: String)` returns an empty string if the config directory doesn't exist. As a result, the first default config created receives an empty save path.

KANAjetzt:
With this change, these functions will always return the full path to the config file/dir, even if the parent directory doesn't exist. Usually, this directory will be created later when the file is saved.

closes #267

### DIFF
--- a/addons/mod_loader/internal/path.gd
+++ b/addons/mod_loader/internal/path.gd
@@ -149,13 +149,11 @@ static func get_path_to_configs() -> String:
 
 
 # Get the path to a mods config folder
-# Returns an empty string if there is no config dir for this mod_id
 static func get_path_to_mod_configs_dir(mod_id: String) -> String:
 	return get_path_to_configs().plus_file(mod_id)
 
 
 # Get the path to a mods config file
-# Returns an empty string if the config file does not exist.
 static func get_path_to_mod_config_file(mod_id: String, config_name: String) -> String:
 	var mod_config_dir := get_path_to_mod_configs_dir(mod_id)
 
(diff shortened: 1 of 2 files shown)


addons/mod_loader/internal/path.gd
```diff
@@ -151,27 +151,12 @@ static func get_path_to_configs() -> String:
 # Get the path to a mods config folder
 # Returns an empty string if there is no config dir for this mod_id
```
Qubus0, line 152: comments aren't up to date

Qubus0 (approved):

addons/mod_loader/internal/path.gd
KANAjetzt: ty - fixed 05ff0a2

PR #270, opened 2023-06-11
fix: :bug: added a check if `current_user_profile` is `null`

### LINKED ISSUE #268: `ModLoaderConfig` - Error Invalid get index 'name' when loading configs
```gdscript
static func get_current_config_name(mod_id: String) -> String:
	# Check if user profile has been loaded
---> 	if not ModLoaderStore.user_profiles.has(ModLoaderStore.current_user_profile.name):
```

image

KANAjetzt:
closes #268


Qubus0 (approved):
better safe than sorry

PR #271, opened 2023-06-11
style: :pencil2: `add_translation_from_resource` to `add_translation`

KANAjetzt:
We removed the old way of adding translations a while back and now only have the `add_translation_from_resource(resource_path: String)` method, so we can simply rename it to `add_translation(resource_path: String)`.


Qubus0 (approved):
✨

PR #274, opened 2023-06-12
fix: :bug: added `get_mod_data_all()`

KANAjetzt:
Added `get_mod_data_all()` and updated the deprecation message to point to `ModLoaderMod.get_mod_data_all()` instead of `ModLoaderMod.get_mod_data()`.


Qubus0 (approved):
yep, better 👍

PR #276, opened 2023-06-13
fix: always use warnings for deprecations in standalone builds

Qubus0:
There seems to be an issue where deprecations can cause mods to stop working even through the `assert` from `log-fatal` should be removed. 
To circumvent this, turn all deprecations into warnings for standalone (non-editor) builds


otDan (approved):
yup

PR #277, opened 2023-06-16
fix: Transitive Comparator for script sorting 

boardengineer:
The current algorithm returns a random result unless a is in b's type hierarchy.  Random results are disallowed in the sort_custom API

https://docs.godotengine.org/en/3.5/classes/class_array.html?highlight=sort_custom

The new algorithm sorts alphabetically by names in the type hierarchy paths until one of the trees stop then it sorts the shorter tree first.

Aims to fix the following errors that gets spammed at startup (and i think causes startup crashes):

E 0:00:00.815   SortArray<class Variant,struct _ArrayVariantSortCustom,1>::partitioner: bad comparison function; sorting will be broken
  <C++ Source>  .\core/sort_array.h:179 @ SortArray<class Variant,struct _ArrayVariantSortCustom,1>::partitioner()
  <Stack Trace> script_extension.gd:39 @ handle_script_extensions()
                mod_loader.gd:164 @ _load_mods()
                mod_loader.gd:67 @ _init()

### DIFF
--- a/addons/mod_loader/internal/script_extension.gd
+++ b/addons/mod_loader/internal/script_extension.gd
@@ -50,12 +50,14 @@ class InheritanceSorting:
 		while parent_script:
 			a_stack.push_front(parent_script.resource_path)
 			parent_script = parent_script.get_base_script()
+		a_stack.pop_back()
 		
 		var b_stack := []
 		parent_script = load(extension_b.extension_path)
 		while parent_script:
 			b_stack.push_front(parent_script.resource_path)
 			parent_script = parent_script.get_base_script()
+		b_stack.pop_back()
 		
 		var last_index: int
 		for index in a_stack.size():
--- a/addons/mod_loader/internal/script_extension.gd
+++ b/addons/mod_loader/internal/script_extension.gd
@@ -54,13 +54,13 @@ class InheritanceSorting:
 		var a_stack = []
 		var parent_script = load(extension_a.extension_path)
 		while parent_script:
-			a_stack.push_back(parent_script.resource_path)
+			a_stack.push_front(parent_script.resource_path)
 			parent_script = parent_script.get_base_script()
 		
 		var b_stack = []
 		parent_script = load(extension_b.extension_path)
 		while parent_script:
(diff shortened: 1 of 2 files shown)


addons/mod_loader/internal/script_extension.gd
```diff
@@ -49,28 +49,33 @@ static func handle_script_extensions() -> void:
...
-		else:
-			a_child_script = ResourceLoader.load(extension_a.parent_script_path)
-			ModLoaderStore.loaded_vanilla_parents_cache[extension_a.parent_script_path] = a_child_script
-
-		var a_parent_script: Script = a_child_script.get_base_script()
-
-		if a_parent_script == null:
+	
+	static func _check_inheritances(extension_a:ScriptExtensionData, extension_b:ScriptExtensionData)->bool:
+		var a_stack = []
```
Qubus0, line 54: please try to use static types -> `:=`
(also applies to the rest of the method)

Qubus0 (commented):
Looks good. Thank you for the contribution 😃 

Comparing the stacks looks wrong at first glance since it just compares the scripts at the same index in the stack, but with `push_front` it starts to make sense. Both stacks start after the native Godot classes and work up from there. No matter where both stacks would go apart, they will always arrive at the same vanilla script at the same time. 
On that note - I assume `get_base_script` does not return Godot classes, right? Otherwise it could cause issues

Another thing: The code does not use `ModLoaderStore.loaded_vanilla_parents_cache.keys()` anymore. I think this was the only place where it was used, so it can be removed from the store and also where it is updated. Please have a look

addons/mod_loader/internal/script_extension.gd
boardengineer: Added static types and removed cache, Thank you

Qubus0 (approved):
nice, thank you.

Qubus0 (changes requested):
blocked until the bug we discussed on discord is fixed

Qubus0 (approved):
important fix, thank you 👍

PR #278, opened 2023-06-16
fix: :bug: log stack only saved for `logged_messages.all`

KANAjetzt:
Previously, the same log entry was only saved to the previous entry in `ModLoaderStore.logged_messages.all[logMD5]`. This change adds a check for an `existing_entry` and adds to the entry stacks in `ModLoaderStore.logged_messages.by_mod` and `ModLoaderStore.logged_messages.by_type`.


Qubus0 (approved):
good catch

PR #293, opened 2023-06-21
fix: :bug: `current_user_profile` reference to `user_profiles`

### LINKED ISSUE #290: ModLoaderUserProfile - Current User Profile bug
The current user profile doesn't match the file state of the current user profile.
Because it is currently a separate instance of what is stored in `ModLoaderStore.user_profiles`

https://clips.twitch.tv/HedonisticJoyousStarlingNononoCat-z7CnsEDeGgVsXmGw
https://clips.twitch.tv/PiercingTiredUdonSwiftRage-W8-agV-oxy9-JO4o

KANAjetzt:
Resolves a bug where the `current_user_profile` doesn't reference the Profile object in `user_profiles`.

closes #290


addons/mod_loader/api/profile.gd
```diff
@@ -80,12 +80,12 @@ static func create_profile(profile_name: String) -> bool:
...
-	ModLoaderStore.current_user_profile = new_profile
-
 	# Store the new profile in the ModLoaderStore
 	ModLoaderStore.user_profiles[profile_name] = new_profile
 
+	# Set it as the current profile
+	ModLoaderStore.current_user_profile = ModLoaderStore.user_profiles[profile_name]
+
 	# Store the new profile in the json file
 	var is_save_success := _save()
```
Qubus0, line 90: should we reset the current user profile if the save failed?

KANAjetzt: Hmm, if the save failed, something must have gone wrong with accessing the file system.

Qubus0: Yes. And that means there will be a different user profile when the game is restarted. Not a thing we have to solve now, but possibly a point of confusion in the future

Qubus0 (approved):

addons/mod_loader/api/profile.gd
KANAjetzt: opened #295 for this 👍

PR #305, opened 2023-06-28
chore: :arrow_up: update version const for v6.0.1

KANAjetzt:


Qubus0 (approved):
that could have been awkward 😬 
maybe we can add a check to validate if that constant is the same as the pr title for any pr tagged as release. might be too much effort though

Issue #306, opened 2023-06-29
ModLoaderUserProfile - `mod_list` is not being updated correctly.


KANAjetzt:
The uninstall detection does not work properly because the `zip_path` is not stored in `mod_user_profiles.json`.

<details><summary>Video</summary>
<p>

https://github.com/GodotModding/godot-mod-loader/assets/41547570/d0202571-4715-4ac8-a71d-39b2af06638f

</p>
</details>

KANAjetzt:
closed by #307

PR #310, opened 2023-06-30
v6.0.2

KANAjetzt:
# Patch v6.0.2

This patch addresses an issues related to the new user profiles introduced in v6.0.0. It also includes an update for the manifest validation.

### Bug fix for `ModLoaderUserProfile`
- #306 
- #307

### Manifest Validation
- #308
- #312

## 🚨 Breaking Changes
- Technically, it may break mods that have the same `mod_id` listed in both `optional_dependencies` and `dependencies` or `load_before` and `incompatibilities`. However, since the `optional_dependencies` and `load_before` feature was introduced in v6.0.0, it is highly unlikely that such conflicts exist.

### DIFF
--- a/addons/mod_loader/api/profile.gd
+++ b/addons/mod_loader/api/profile.gd
@@ -253,17 +253,8 @@ static func _update_mod_list(mod_list: Dictionary, mod_data := ModLoaderStore.mo
 	var updated_mod_list := mod_list.duplicate(true)
 
 	# Iterate over each mod ID in the mod list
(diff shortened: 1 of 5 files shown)


ithinkandicode (commented):
Don't forget the version bump in mod_loader_store!

https://github.com/GodotModding/godot-mod-loader/blob/279a012ff9d142ad2b9b94f76e8a70d7f75ba746/addons/mod_loader/mod_loader_store.gd#L15

Qubus0 (approved):

PR #313, opened 2023-06-30
Sort cache

boardengineer:


addons/mod_loader/internal/script_extension.gd
```diff
@@ -115,7 +104,7 @@ static func apply_extension(extension_path: String) -> Script:
 	return child_script
 
 
-# Sort an array of ScriptExtensionData following the load order
+# TODO: this isn't being called
```
otDan, line 107: is there a reason why you left it out instead of deleting it then?

otDan:
also could you rename the pr to something like `feat: Cache for extension sorting`

addons/mod_loader/internal/script_extension.gd
boardengineer: From context I imagine someone might want this back even though the idea didn't quite work

addons/mod_loader/internal/script_extension.gd
```diff
@@ -7,57 +7,31 @@ extends Reference
...
-	for extension in script_extension_data_array:
-		var script: Script = apply_extension(extension.extension_path)
+	for extension in extension_paths:
+		var script: Script = apply_extension(extension)
 		_reload_vanilla_child_classes_for(script)
 
 
 # Inner class so the sort function can be called by handle_script_extensions()
 class InheritanceSorting:
+	var stack_cache = {}
```
otDan, line 30: also I am confused, wasn't this supposed to be a file cache that helps load up faster if we don't change the mods? this seems to only work at runtime or am I missing something?

PR #314, opened 2023-06-30
feat: Implement a cache of script inheritence

### LINKED ISSUE #311: Caching sorted script extensions and mod load order
Currently, two of the heaviest parts of loading mods are being executed even when there are no changes to the installed mods. This results in unnecessary processing and delays in mod loading.

To improve performance, I suggest implementing a caching mechanism for the sorted script extensions and mod load order.

## Load Order
https://github.com/GodotModding/godot-mod-loader/blob/6c91c475c0a9bbae323ca43bebdbee2a0fdfc666/addons/mod_loader/mod_loader.gd#L154-L155

## Script Extensions
https://github.com/GodotModding/godot-mod-loader/blob/6c91c475c0a9bbae323ca43bebdbee2a0fdfc666/addons/mod_loader/internal/script_extension.gd#L11-L41

boardengineer:
There was a clobbered importance order that is no longer being called cosmetically

the Inheritance stacks are now being saved so they don't have to be recalced again on each iteration, this saves the majority of the sorting time

A ScriptExtensionData type has been removed since it was only used to keep attributes together through the sort and a bit of unnecessary work was done to maintain the type. 

<br />

closes
#311


addons/mod_loader/internal/script_extension.gd
```diff
@@ -115,7 +104,7 @@ static func apply_extension(extension_path: String) -> Script:
 	return child_script
 
 
-# Sort an array of ScriptExtensionData following the load order
+# TODO: this isn't being called
```
KANAjetzt, line 107: What is todo here - additional caching? 

I can see the call for `_sort_extensions_from_load_order()` has been removed in `handle_script_extensions()` so is it no longer required?

addons/mod_loader/internal/script_extension.gd
```diff
@@ -71,7 +45,22 @@ class InheritanceSorting:
 			# 'a' has a shorter stack
 			return true
 
-		return extension_a.extension_path < extension_b.extension_path
+		return extension_a < extension_b
+	
+	func cached_stack(extension_path:String) -> Array:
```
KANAjetzt, line 50: Small style thing:

`func cached_stack(extension_path:String) -> Array:` -> `func cached_stack(extension_path: String) -> Array:`

addons/mod_loader/internal/script_extension.gd
```diff
@@ -7,57 +7,31 @@ extends Reference
...
-	for extension in script_extension_data_array:
-		var script: Script = apply_extension(extension.extension_path)
+	for extension in extension_paths:
+		var script: Script = apply_extension(extension)
 		_reload_vanilla_child_classes_for(script)
 
 
 # Inner class so the sort function can be called by handle_script_extensions()
 class InheritanceSorting:
+	var stack_cache = {}
```
KANAjetzt, line 30: Type
`var stack_cache = {}` -> `var stack_cache := {}`

addons/mod_loader/internal/script_extension.gd
KANAjetzt, line 50: Can you add some comments to explain what this code does? Together with `_check_inheritances()`

addons/mod_loader/resources/script_extension_data.gd
```diff
@@ -1,20 +0,0 @@
-class_name ScriptExtensionData
```
KANAjetzt, line 1: Can you add to the PR description why this `Resource` can be removed?

KANAjetzt (commented):
Thank You!
I tested it with my 18 mods in Brotato, and the difference is like night and day.
( And still works as expected 👀 )

addons/mod_loader/internal/script_extension.gd
boardengineer: The todo is to (starts with) rethink this feature since just sorting twice doesn't work.  I left it as a tombstone since it was silently turned into a no-op when the second sort was introduced.

addons/mod_loader/resources/script_extension_data.gd
boardengineer: I honestly can't figure out how to edit a "PR description"

KANAjetzt (approved):
Nice! Thanks again for your help. Feel free to merge.

PR #324, opened 2023-07-03
refactor: Remove self argument on *mod_main.gd* _init

otDan:


otDan:
also let me know a cleared way to tell the user to not use the argument as it's deprecated, currently it doesn't look right

addons/mod_loader/mod_loader.gd
```diff
@@ -336,7 +336,18 @@ func _init_mod(mod: ModData) -> void:
 	var mod_main_script := ResourceLoader.load(mod_main_path)
```
Qubus0, line 336: to improve typing for the script here, I think we should explicitly type to `GDScript`, since load is typed as Resource

addons/mod_loader/mod_loader.gd
```diff
@@ -336,7 +336,18 @@ func _init_mod(mod: ModData) -> void:
...
+	var argument_found: bool = false
+	for method in mod_main_script.get_script_method_list():
+		if method.name == "_init":
+			if method.args.size() > 0:
+				argument_found = true
+	
+	var mod_main_instance: Node
+	if argument_found:
+		mod_main_instance = mod_main_script.new(self)
+		ModLoaderDeprecated.deprecated_changed("ModLoader _init argument", "No ModLoader _init argument", "6.1.0")
```
Qubus0, line 348:
```gdscript
deprecated_message("The mod_main.gd _init argument (modLoader = ModLoader) is deprecated. Remove it from your _init to avoid crashes in the next major version.", "6.1.0")
```

Qubus0 (approved):
yep that is safer

PR #326, opened 2023-07-07
feat: add dictionary for overrides based on feature tags to options

### LINKED ISSUE #296: Production and Development Environment Options
> René — 20/06/2023 22:03
change #1 res://addons/mod_loader/mod_loader_store.gd
I don't need the options.tres, but want two option profiles: one for development, one for release. Release should be automatically used, so i can't forget to switch it out
>
> ```gdscript
>  func _update_ml_options_from_options_resource() -> void:
>     if OS.has_feature("editor"):
>         var current_options: Resource = preload("res://addons/mod_loader/options/profiles/current.tres")
>         for key in ml_options:
>             ml_options[key] = current_options[key]
>     else:
>         var current_options: Resource = preload("res://addons/mod_loader/options/profiles/production_workshop.tres")
>         for key in ml_options:
>             ml_options[key] = current_options[key]
> ```

### Todos:
- [ ] Implement Feature Flags 

https://github.com/GodotModding/godot-mod-loader/issues/296#issuecomment-1618372594
> some games have different export targets (Dome Keeper: Steam, GOG). To support different options for that case, I think we need the full feature tag system, since we can't predict this. In this case, the gog tag would apply a options file with steam workshop disabled. https://docs.godotengine.org/en/3.5/tutorials/export/feature_tags.html#custom-features

Qubus0:
adds a dictionary for overrides based on feature tags to options.

I had to modify ModLoaderLog a little to avoid crashes when the store - and the options themselves in the store - are not available yet, which is the case in the options loading method.

needs to be documented on the options page after merge to main
```
# Overrides for all available feature tags through OS.has_feature()
# Format: Dictionary[String: ModLoaderOptionsProfile] where the string is a tag
# Warning: Some tags can occur at the same time (Windows + editor for example) -
# In a case where multiple apply, the last one in the dict will override all others
```

https://docs.godotengine.org/en/3.5/tutorials/export/feature_tags.html

closes #296

### DIFF
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -345,7 +345,8 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 	if only_once and _is_logged_before(log_entry):
 		return
 
-	_store_log(log_entry)
+	if ModLoaderStore:
+		_store_log(log_entry)
 
 	# Check if the scene_tree is available
 	if Engine.get_main_loop():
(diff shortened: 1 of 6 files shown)


KANAjetzt (approved):
Nice - Thank You! 
Tested it in Editor with `windows` and `editor` flag.

PR #331, opened 2023-07-17
feat: log name wildcards

Qubus0:
Before, it was only possible to match 100% of a log name to ignore it e.g. `ModLoader:Deprecated`. This can get tedious if you want to ignore all logs from a log name. 
Now you can use a wildcard at the end e.g. `ModLoader:*` to ignore all log names starting with that segment.

### DIFF
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -402,7 +402,7 @@ static func _is_mod_name_ignored(mod_log_name: String) -> bool:
 	# Match a mod log name with a wildcard. ex: "ModLoader:*"
 	for ignored_mod_name in ignored_mod_names:
 		if ignored_mod_name.ends_with("*"):
-			if mod_log_name.begins_with(ignored_mod_name.replace("*", "")):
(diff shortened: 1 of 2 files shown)


addons/mod_loader/api/log.gd
```diff
@@ -385,15 +385,27 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
...
 		return false
 
-	var ignored_mod_names := ModLoaderStore.ml_options.ignored_mod_names_in_log as Array
+	var ignored_mod_log_names := ModLoaderStore.ml_options.ignored_mod_names_in_log as Array
 
-	if not ignored_mod_names.size() == 0:
-		if mod_name in ignored_mod_names:
-			return true
+	# No ignored mod names
+	if ignored_mod_names.size() == 0:
```
KANAjetzt, line 395: `ignored_mod_names` should be `ignored_mod_log_names` now - I think.

Qubus0: yup

KANAjetzt (approved):
ty 👍

PR #332, opened 2023-08-24
Added debugLine() method

aesereht:
Script pointing the top stack to get script name, function and line.

### DIFF
--- a/addons/mod_loader/api/deprecated.gd
+++ b/addons/mod_loader/api/deprecated.gd
@@ -65,3 +65,4 @@ static func _deprecated_log(msg: String) -> void:
 		ModLoaderLog.warning(msg, LOG_NAME)
 	else:
 		ModLoaderLog.fatal(msg, LOG_NAME)
+	ModLoaderLog.debugLine()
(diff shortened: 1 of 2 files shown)


KANAjetzt:
The logs can be sometimes a bit confusing, but in most cases its very helpful.

Not so helpful:
```
WARNING ModLoader:Deprecated: DEPRECATED: The mod_main.gd _init argument (modLoader = ModLoader) is deprecated. Remove it from your _init to avoid crashes in the next major version. (since version 6.1.0)
Problem File - Source:res://addons/mod_loader/mod_loader.gd - Function:_init - Line:65
```

Helpful:
```
WARNING ModLoader:Deprecated: DEPRECATED: The var "mod_data" was removed, use "ModLoaderMod.get_mod_data_all()" instead (since version 6.0.0)
Problem File - Source:res://mods-unpacked/otDan-BetterModList/extensions/ui/menus/title_screen/title_screen_menus.gd - Function:_ready - Line:7
```


Would be nice if we could move the origin of the deprecation warning to the top of the debugger stack.
image

So the mod dev only gets the deprecation error message and the origin in the debugger.

PR #333, opened 2023-08-25
4.x branch update to 6.0.2 version

MateuSai:
I checked the merge changes of versions 6.0.1 and 6.0.2 and I made the necessary changes to the 4.x branch

### DIFF
--- a/addons/JSON_Schema_Validator/json_schema_validator.gd
+++ b/addons/JSON_Schema_Validator/json_schema_validator.gd
@@ -88,18 +88,18 @@ const ERR_WRONG_PATTERN = "String '%s' does not match its corresponding pattern"
 const ERR_FORMAT = "String '%s' does not match its corresponding format '%s'"
 
 # This is one and only function that need you to call outside
-# If all validation checks passes, this return empty String
+# If all validation checks passes, this OK
 func validate(json_data : String, schema: String) -> String:
-	var error: String
+	var error: int
 
 	var json = JSON.new()
 	# General validation input data as JSON file
-	error = str(json.parse(json_data))
-	if error: return ERR_INVALID_JSON_EXT % error
+	error = json.parse(json_data)
+	if error: return ERR_INVALID_JSON_EXT % error_string(error)
 
 	# General validation input schema as JSONSchema file
-	error = str(json.parse(schema))
-	if not error == "OK" : return ERR_WRONG_SCHEMA_GEN + str(error)
+	error = json.parse(schema)
+	if not error == OK : return ERR_WRONG_SCHEMA_GEN + error_string(error)
 	var test_json_conv = JSON.new()
 	test_json_conv.parse(schema)
 	var parsed_schema = test_json_conv.get_data()
@@ -117,10 +117,8 @@ func validate(json_data : String, schema: String) -> String:
 		_: return ERR_WRONG_SCHEMA_TYPE
 
 	# All inputs seems valid. Begin type validation
-	error = _type_selection(json_data, parsed_schema)
-
 	# Normal return empty string, meaning OK
-	return error
+	return _type_selection(json_data, parsed_schema)
 
 func _to_string():
 	return "[JSONSchema:%d]" % get_instance_id()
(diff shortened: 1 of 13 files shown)


MateuSai:
Let me explain what I did on the last commit. I was getting this error:

"FATAL-ERROR ModLoader:File: Encountered an error when attempting to open a file, with the path: user://mod_user_profiles.json"

So, I checked the `_save_string_to_file` function and I found this:

```swift
static func _save_string_to_file(save_string: String, filepath: String) -> bool:
	# Create directory if it doesn't exist yet
	var file_directory := filepath.get_base_dir()
	var dir := DirAccess.open(file_directory)

	_code_note(str(
		"View error codes here:",
		"https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-error"
	))

	if not dir.dir_exists(file_directory):
		var makedir_error := dir.make_dir_recursive(file_directory)
		if not makedir_error == OK:
			ModLoaderLog.fatal("Encountered an error (%s) when attempting to create a directory, with the path: %s" % [makedir_error, file_directory], LOG_NAME)
			return false

	if file_exists(filepath):
		ModLoaderLog.fatal("Encountered an error when attempting to open a file, with the path: %s" % [filepath], LOG_NAME)
		return false

	# Save data to the file
	var file := FileAccess.open(filepath, FileAccess.WRITE)

	file.store_string(save_string)
	file.close()

	return true
```

It's throwing an error if the file already exists, which I think is wrong since user://mod_user_profiles.json should be overwritten. I checked the same function on the main branch and it's like this:

```swift
static func _save_string_to_file(save_string: String, filepath: String) -> bool:
	# Create directory if it doesn't exist yet
	var file_directory := filepath.get_base_dir()
	var dir := Directory.new()

	_code_note(str(
		"View error codes here:",
		"https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-error"
	))

	if not dir.dir_exists(file_directory):
		var makedir_error := dir.make_dir_recursive(file_directory)
		if not makedir_error == OK:
			ModLoaderLog.fatal("Encountered an error (%s) when attempting to create a directory, with the path: %s" % [makedir_error, file_directory], LOG_NAME)
			return false

	var file := File.new()

	# Save data to the file
	var fileopen_error := file.open(filepath, File.WRITE)

	if not fileopen_error == OK:
		ModLoaderLog.fatal("Encountered an error (%s) when attempting to write to a file, with the path: %s" % [fileopen_error, filepath], LOG_NAME)
		return false

	file.store_string(save_string)
	file.close()

	return true

```

As you see, here it throws an error if it can't open the file, not if the file already exists. So I modified the code on the 4.x branch so that it does the same:

```swift
static func _save_string_to_file(save_string: String, filepath: String) -> bool:
	# Create directory if it doesn't exist yet
	var file_directory := filepath.get_base_dir()
	var dir := DirAccess.open(file_directory)

	_code_note(str(
		"View error codes here:",
		"https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-error"
	))

	if not dir.dir_exists(file_directory):
		var makedir_error := dir.make_dir_recursive(file_directory)
		if not makedir_error == OK:
			ModLoaderLog.fatal("Encountered an error (%s) when attempting to create a directory, with the path: %s" % [makedir_error, file_directory], LOG_NAME)
			return false

	# Save data to the file
	var file := FileAccess.open(filepath, FileAccess.WRITE)

	if not file:
		ModLoaderLog.fatal("Encountered an error when attempting to open a file, with the path: %s" % [filepath], LOG_NAME)
		return false

	file.store_string(save_string)
	file.close()

	return true
```

otDan:
I was holding off from porting latest since no one tested 4.x in general

KANAjetzt:
@MateuSai is testing it now, what do you think is the best way to update the 4.x branch to 6.2.0 @otDan ?
Looks like the first issue hit, are the bugs we fixed in the later versions.

KANAjetzt (approved):
Nice work - Thank You! 🥇

PR #363, opened 2024-01-24
fix: non-static func call on `InheritanceSorting`

MateuSai:
At the line 22 of `script_extension.gd`, there is this line:

`extension_paths.sort_custom(InheritanceSorting._check_inheritances)` (I adapted it to Godot 4 by using the function reference instead of using Callable(...))

Since we don't create an instance of `InheritanceSorting`, _check_inheritances must be static, otherwise we get errors:

image

The variable `stack_cache` and the function `cached_inheritances_stack` of the same class, are used on `_check_inheritances`, so they need to be static too.

### DIFF
--- a/addons/mod_loader/internal/script_extension.gd
+++ b/addons/mod_loader/internal/script_extension.gd
@@ -19,7 +19,7 @@ static func handle_script_extensions() -> void:
 			)
 
 		# Sort by inheritance
-		extension_paths.sort_custom(Callable(InheritanceSorting, "_check_inheritances"))
+		extension_paths.sort_custom(InheritanceSorting._check_inheritances)
 
 		# Load and install all extensions
 		for extension in extension_paths:
@@ -31,11 +31,11 @@ static func handle_script_extensions() -> void:
 # ancestors then sorted such that scripts extending script A will be before
 # a script extending script B if A is an ancestor of B.
 class InheritanceSorting:
-	var stack_cache := {}
+	static var stack_cache := {}
 
 	# Comparator function.  return true if a should go before b.  This may
 	# enforce conditions beyond the stated inheritance relationship.
-	func _check_inheritances(extension_a: String, extension_b: String) -> bool:
+	static func _check_inheritances(extension_a: String, extension_b: String) -> bool:
 		var a_stack := cached_inheritances_stack(extension_a)
 		var b_stack := cached_inheritances_stack(extension_b)
 
@@ -56,7 +56,7 @@ class InheritanceSorting:
 	# script with the most recent ancestor last.
 	#
 	# Results are stored in a cache keyed by extension path
-	func cached_inheritances_stack(extension_path: String) -> Array:
+	static func cached_inheritances_stack(extension_path: String) -> Array:
 		if stack_cache.has(extension_path):
 			return stack_cache[extension_path]
 
@@ -239,3 +239,4 @@ static func remove_all_extensions_of_mod(mod: ModData) -> void:
 	for extension_path in _to_remove_extension_paths:
 		remove_specific_extension_from_script(extension_path)
 		ModLoaderStore.saved_extension_paths.erase(mod.manifest.get_mod_id())
+
--- a/addons/mod_loader/internal/script_extension.gd
+++ b/addons/mod_loader/internal/script_extension.gd
@@ -19,8 +19,7 @@ static func handle_script_extensions() -> void:
 			)
 
 	# Sort by inheritance
-	var new_inheritance_sorting := InheritanceSorting.new()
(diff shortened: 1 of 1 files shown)


KANAjetzt:
Thanks for your PR! 
Can you provide the Godot version and perhaps two test mods where you encountered this error?

MateuSai:
I'm using Godot 4.2.1, but I don't think it's a bug related with the Godot version. The functions you call from a static function must be static too. If you replace `Callable(...)` with `InheritanceSorting._check_inheritances`, an error appears automatically on the code editor telling you about this. With `Callable(InheritanceSorting, "_check_inheritances")` you don't notice an error until you execute that code.

The error appear after extending a script with `ModLoaderMod.install_script_extension(...)`. It does not create a crash, so maybe that's why it has been unnoticed until now.

I can make a basic project with a mod that extends a script for you to see the error if you want.

KANAjetzt:
I currently don't have a solid test environment, so if you have something, that would be very helpful 👍

MateuSai:
Here you have it.

[ModLoader script extension bug.zip](https://github.com/GodotModding/godot-mod-loader/files/14063476/ModLoader.script.extension.bug.zip)

`extension_paths.sort_custom(Callable(InheritanceSorting, "_check_inheritances"))` is executing and it does not generate the errors I had seen. However, if you put a breakpoint in the `_check_inheritances` function, it won't be reached. So, I assume Godot is not showing you any error but the function is not being executed? Otherwise I can't understand.

You might also want to check the line 352 of `log.gd`. With `if Engine.get_main_loop():` it crashes at startup. `if ModLoader:` seems to work as a replacement.

boardengineer:
my 2 cents:

This is the wrong approach, you should be trying to make fewer things static instead of trying to make more things static

addons/mod_loader/internal/script_extension.gd
```diff
@@ -31,11 +31,11 @@ static func handle_script_extensions() -> void:
 # ancestors then sorted such that scripts extending script A will be before
 # a script extending script B if A is an ancestor of B.
 class InheritanceSorting:
-	var stack_cache := {}
+	static var stack_cache := {}
```
boardengineer, line 34: Can you just instantiate this class instead?

ITR13: Can confirm that the following also works:
```gdscript
# Sort script extensions by inheritance and apply them in order
static func handle_script_extensions() -> void:
	var extension_paths := []
	for extension_path in ModLoaderStore.script_extensions:
		if FileAccess.file_exists(extension_path):
			extension_paths.push_back(extension_path)
		else:
			ModLoaderLog.error(
				"The child script path '%s' does not exist" % [extension_path], LOG_NAME
			)

		# Sort by inheritance
	        var sorter = InheritanceSorting.new()
		extension_paths.sort_custom(Callable(sorter, "_check_inheritances"))

		# Load and install all extensions
		for extension in extension_paths:
			var script: Script = apply_extension(extension)
```

Will also note that it's probably indented wrong? Surely it's not intended to sort extension_paths and recall apply_extension every single time it adds an extension?

KANAjetzt: done in [b4c37a7](https://github.com/GodotModding/godot-mod-loader/pull/363/commits/b4c37a79ad09a393bb992fc3e0f76aaa74a2f335) 👍 

Thanks vor reviewing and the PR!

addons/mod_loader/internal/script_extension.gd
```diff
@@ -19,7 +19,8 @@ static func handle_script_extensions() -> void:
 			)
 
 	# Sort by inheritance
-	extension_paths.sort_custom(Callable(InheritanceSorting, "_check_inheritances"))
+	var new_inheritance_sorting := InheritanceSorting.new()
+	extension_paths.sort_custom(new_inheritance_sorting.check_inheritances)
```
Qubus0, line 23: just a quick idea, if we call the method in the constructor of `InheritanceSorting`, we can keep everything concise in one line. the constructor does not return anything, of course, but since the array is a reference we can still sort this way.

minimal example:

```gdscript
var extension_paths := ["d", "b", "c", "f", "a", "e"]

func _ready():
	print(extension_paths)
	InheritanceSorting.new(extension_paths)
	print(extension_paths)

class InheritanceSorting:
	func _init(array: Array) -> void:
		array.sort_custom(check_inheritances)

	func check_inheritances(extension_a: String, extension_b: String) -> bool:
		return extension_a < extension_b
```

KANAjetzt: beautiful ✨ - done in eb0873d

Qubus0 (approved):

Issue #353, opened 2023-10-28
Possible memory leak with script extensions


KANAjetzt:
KoBeWi brought to the attention that this line:

https://github.com/GodotModding/godot-mod-loader/blob/19a7b7fc484a98969d9a82ff8f402a3a30f08246/addons/mod_loader/internal/script_extension.gd#L93

Can cause a potential memory leak if the `child_script` does not inherit from [RefCounted](https://docs.godotengine.org/en/stable/classes/class_refcounted.html) (or [Reference](https://docs.godotengine.org/en/3.5/classes/class_reference.html) in 3.x).

They suggested using `.reload()` instead.

KANAjetzt:
fixed for 4.x with https://github.com/GodotModding/godot-mod-loader/pull/383

KANAjetzt:
fixed for 3.x in #522

Issue #338, opened 2023-09-24
Several mods cannot extend the same script with a `class_name`


ColdCalzone:
In Godot 4, having two or more mods extend the same script causes every script after the first to be ignored. I've found that `parent_script.resource_path` is blank after the first extension takes over.
After modifying the `apply_extension` function to sidestep that issue, I found that every extension except the last one is ignored.

ColdCalzone:
I've discovered the problem isn't with *every* script, it only happens when the script is declared with a `class_name` at the top, or extends a script declared with a `class_name`. Autoloads and regular scripts can be extended just fine.

KANAjetzt:
I'm currently testing some older Godot version and got this interesting error message on 4.0-RC1
![image](https://github.com/GodotModding/godot-mod-loader/assets/41547570/4be16ceb-1cb3-4281-b07f-8ec78b529e05)


https://github.com/godotengine/godot/commits/master/core/io/resource_loader.cpp

KANAjetzt:
Scripts with `class_name` still can't be extended using `take_over_path` / script extensions. As a workaround, we introduced Mod Hooks in #408. More details can be found on the [Wiki](https://wiki.godotmodding.com/guides/modding/script_hooks/).  

Closed by #408

PR #365, opened 2024-03-10
refactor: Rewrite applying extensions

cgytrus:
Fixes the currently main 4.x issue
- disallows installing extensions outside of _init
  - same reason as below, allowing this would require restructuring my code again..
- removes runtime mod toggling/reloading
  - too complicated for me to implement with the rewrite and ive been told it was unstable anyway
- fixes preloaded scenes that refer to extended scripts not getting updated
  - if you have a scene that refers to a script, that scene gets preloaded and the script gets extended, the preloaded version of the scene would still refer to the old script, i believe the use of `reload` instead of `new` fixes that but i havent actually checked if thats true
- the rewrite ends up fixing these
  - fixes https://github.com/GodotModding/godot-mod-loader/issues/338
  - fixes https://github.com/GodotModding/godot-mod-loader/issues/353
- ~~supersedes https://github.com/GodotModding/godot-mod-loader/pull/363~~ (already merged anyway)

how this works:
- collects all extension scripts (no changes here)
- uses the base script as the original script that will be extended (no changes here either)
- copies the original script to a temporary directory on disk
  - it seems to be impossible to create a new file in the project otherwise and `extends` ignores resource cache, which is what `take_over_path` overrides, it reads only from packs or the file system
- `extends` in the extension scripts is patched to instead extend the script previous in the mod hierarchy, down to the *copied* original file
- the original script's source is overwritten with the source of the last applied extension
  - `take_over_path` didnt seem to work for this
  - the hierarchy ends up as `mod (overwritten original) > mod > mod > ... > copied original`
- all involved scripts are reloaded in backwards order (starting from copied original, ending with overwritten original)
  - this ensures that all the scripts are up to date with the patches and preloads and whatnot
- temporary directory together with all the patched extensions and copied original is deleted from disk as theyre now loaded and not needed anymore

i didnt test much, especially with dependencies, but for a simple use case of a couple mods this seems to be working
also please squash when merging lol

current issues:
 - [ ] there seems to be an issue where this also ends up hijacking the paths in the debugger too, when you click on an extended script in a stack trace it opens files that were either already deleted from the temp folder or original files that had their source code replaced in runtime


Qubus0:
thanks for your contributions again. this was a good experiment and we learned quite a few things from it

PR #368, opened 2024-03-11
fix: config MD5 check

cgytrus:
the condition here was incorrect, causing configs to always get overwritten on startup, this commit fixes the issue, the comment above confirms that this was the intended condition too

### DIFF
--- a/addons/mod_loader/resources/mod_manifest.gd
+++ b/addons/mod_loader/resources/mod_manifest.gd
@@ -227,7 +227,7 @@ func load_mod_config_defaults() -> ModConfig:
 		var cache_schema_md5: String = cache_schema_md5s[config.mod_id] if cache_schema_md5s.has(config.mod_id) else ''
 
 		# Generate a new default config if the config schema has changed or there is nothing cached
-		if not current_schema_md5 == cache_schema_md5 or not cache_schema_md5.is_empty():
+		if current_schema_md5 != cache_schema_md5 or cache_schema_md5.is_empty():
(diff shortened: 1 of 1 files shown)


addons/mod_loader/resources/mod_manifest.gd
```diff
@@ -227,7 +227,7 @@ func load_mod_config_defaults() -> ModConfig:
 		var cache_schema_md5: String = cache_schema_md5s[config.mod_id] if cache_schema_md5s.has(config.mod_id) else ''
 
 		# Generate a new default config if the config schema has changed or there is nothing cached
-		if not current_schema_md5 == cache_schema_md5 or not cache_schema_md5.is_empty():
+		if current_schema_md5 != cache_schema_md5 or cache_schema_md5.is_empty():
```
KANAjetzt, line 230: I think @Qubus0 mentioned `!=` is not best practice? 

```suggestion
		if not current_schema_md5 == cache_schema_md5 or cache_schema_md5.is_empty():
```

cgytrus: sorry, im used to prefer that coming from other languages, if the code style in your project is to specifically use `not ==` instead of `!=`, i can change that back

KANAjetzt: Relatable, but yes, please change it to `not ==`

cgytrus: done!

KANAjetzt (approved):
👍

PR #369, opened 2024-03-11
feat: Allow both workshop and local mods to be loaded

cgytrus:
allows loading mods from both the local mods folder and from the steam workshop folder if workshop is enabled


Qubus0 (approved):
clean, thanks 👍

PR #370, opened 2024-03-11
fix: startup crash in log.gd

cgytrus:
this fix was mentioned in https://github.com/GodotModding/godot-mod-loader/pull/363#issuecomment-1911850304


KANAjetzt (approved):
ty 👍

PR #374, opened 2024-03-21
feat: :sparkles: added `refresh_config_data()` and `refresh_current_c…

KANAjetzt:
Allows refreshing of config data at any time during runtime.

Adds 2 new functions:

**`func refresh_config_data(config: ModConfig) -> ModConfig:`**
Refreshes the data of the provided configuration by reloading it from the config file.

**`func refresh_current_configs() -> void:`**
Iterates over all mods to refresh the data of their current configurations, if available.
Compares the previous configuration data with the refreshed data and emits the `current_config_changed` signal if changes are detected.

### DIFF
--- a/addons/mod_loader/api/config.gd
+++ b/addons/mod_loader/api/config.gd
@@ -345,3 +345,42 @@ static func get_current_config_name(mod_id: String) -> String:
 
 	# Return the name of the current configuration for the mod
 	return current_user_profile.mod_list[mod_id].current_config
+
+
+# Refreshes the data of the provided configuration by reloading it from the config file.
+#
+# Parameters:
+# - config (ModConfig): The ModConfig object whose data needs to be refreshed.
+#
+# Returns:
+# - ModConfig: The ModConfig object with refreshed data if successful, or the original object otherwise.
+static func refresh_config_data(config: ModConfig) -> ModConfig:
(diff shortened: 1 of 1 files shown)


addons/mod_loader/api/config.gd
```diff
@@ -345,3 +345,42 @@ static func get_current_config_name(mod_id: String) -> String:
...
+			continue
+		# Create a deep copy of the current configuration data for comparison
+		var config_data_previous := config.data.duplicate(true)
+		# Refresh the configuration data
+		var config_new := refresh_config_data(config)
+
+		# Compare previous data with refreshed data
+		if not config_data_previous == config_new.data:
+			# Emit signal indicating that the current configuration has changed
+			ModLoader.current_config_changed.emit(config)
```
Qubus0, line 386: this signal is missing from the pr. I'd also add the mod id to the signal for clean access

Qubus0 (changes requested):

addons/mod_loader/api/config.gd
KANAjetzt: the `current_config_changed` signal?

Qubus0: oh that's an existing one. and it also has the id. ignore this then

Qubus0 (approved):

PR #375, opened 2024-03-22
feat: :sparkles: add `has_current_config()` and `has_config()`

KANAjetzt:
Can be used to check if a mod has a config, useful to prevent error prints from the get functions.

### DIFF
--- a/addons/mod_loader/api/config.gd
+++ b/addons/mod_loader/api/config.gd
@@ -295,11 +295,7 @@ static func get_config(mod_id: String, config_name: String) -> ModConfig:
 # - bool: True if the mod has a current configuration, False otherwise.
 static func has_current_config(mod_id: String) -> bool:
 	var mod_data := ModLoaderMod.get_mod_data(mod_id)
-
-	if mod_data.current_config == null:
-		return false
-
-	return true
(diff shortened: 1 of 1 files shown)


addons/mod_loader/api/config.gd
```diff
@@ -284,6 +284,37 @@ static func get_config(mod_id: String, config_name: String) -> ModConfig:
...
+#
+# Returns:
+# - bool: True if the mod has a current configuration, False otherwise.
+static func has_current_config(mod_id: String) -> bool:
+	var mod_data := ModLoaderMod.get_mod_data(mod_id)
+
+	if mod_data.current_config == null:
+		return false
+
+	return true
```
Qubus0, line 302: `return not mod_data.current_config == null`

Qubus0 (approved):

addons/mod_loader/api/config.gd
KANAjetzt: nice - [4a6dcc7](https://github.com/GodotModding/godot-mod-loader/pull/375/commits/4a6dcc7a2561775750c7bf537bf91735f4144e62)

PR #378, opened 2024-04-01
feat: :sparkles: adds `refresh_scene` and `extend_scene`

### LINKED ISSUE #377: Extending scripts attached to a preloaded scene
If a script is attached to a scene that is preloaded in a different script that is also extended the extension for the script of the preloaded scene will not work. This is because the extension scripts are parsed before `take_over_path` is used to add the modified version to the cache.

image

KANAjetzt:
### New Methods

**`static func refresh_scene(scene_path: String) -> void:`**
Refreshes a specific scene by marking it for refresh.

This function is useful if a script extension is not automatically applied. This situation can occur when a script is attached to a preloaded scene. If you encounter issues where your script extension is not working as expected, try to identify the scene to which it is attached and use this method to refresh it. This will reload already loaded scenes and apply the script extension.

🚨 `refresh_scene` will only work for Godot 4.3 and up. 🚨

**`static func extend_scene(scene_vanilla_path: String, edit_callable: Callable) -> void:`**
Extends a specific scene by providing a callable function to modify it. The callable receives an instance of the vanilla_scene as the first parameter.

### Removed Methods

**`static func append_node_in_scene(modified_scene: Node, node_name: String = "", node_parent = null, instance_path: String = "", is_visible: bool = true) -> void:`**
Removed without replacement because it was always unintuitive to use. Instead of this method, the scene extension setup can be used to add a node to a scene.

**`static func save_scene(modified_scene: Node, scene_path: String) -> void:`**
`save_scene` has been moved to *scene_extension.gd* as an internal method used by `handle_scene_extensions()`

## New Class - `_ModLoaderSceneExtension`

Similar to `_ModLoaderScriptExtension`, this class encompasses everything related to modifying PackedScenes, with the hope of continuous improvement.

Currently, preloaded scenes can't be modified using these methods because `CACHE_MODE_REPLACE` loads the scene from file and, with that, wipes any changes stored in the cached instance. For loaded scenes, this should work just fine.

*💡 Preloaded scenes can be modified by adding the code to the script extension.*

## Changes to ModLoader

- Removed deprecation of `append_node_in_scene` and `save_scene`. I think we can remove the other deprecations in a subsequent PR.
- Added calls to `_ModLoaderSceneExtension.refresh_scenes()` and `_ModLoaderSceneExtension.handle_scene_extensions()` at the end of `_load_mods()`

## Changes to ModLoaderStore

- Added `var scenes_to_refresh := []` 
Stores scene paths that need to be reloaded from file.
- Added `var scenes_to_modify := {}` 
Dictionary of callables to modify a specific scene. 
Example property: `"scene_path": [Callable, Callable]`

## Example mod_main.gd
```gdscript
extends Node

const TEST_MOD0_DIR := "Test-Mod0"
const TEST_MOD0_LOG_NAME := "Test-Mod0:Main"

var mod_dir_path := ""
var extensions_dir_path := ""
var translations_dir_path := ""


func _init() -> void:
	mod_dir_path = ModLoaderMod.get_unpacked_dir().path_join(TEST_MOD0_DIR)
	# Add extensions
	install_script_extensions()
	# Edit Scenes
	edit_scenes()


func install_script_extensions() -> void:
	extensions_dir_path = mod_dir_path.path_join("extensions")
	ModLoaderMod.install_script_extension(extensions_dir_path.path_join("preload_scene.gd"))
	ModLoaderMod.install_script_extension(extensions_dir_path.path_join("main.gd"))


func edit_main_scene(scene_instance: Node) -> Node:
	var button: Button = scene_instance.get_node("CanvasLayer/Button")
	button.text = "MODDDED!!!!!"

	return scene_instance


func edit_scenes() -> void:
	ModLoaderMod.refresh_scene("res://preload_scene.tscn")
	ModLoaderMod.extend_scene("res://main.tscn", edit_main_scene)
```

Closes #377

### DIFF
--- a/addons/mod_loader/api/mod.gd
+++ b/addons/mod_loader/api/mod.gd
@@ -95,7 +95,7 @@ static func refresh_scene(scene_path: String) -> void:
 
 
 # Extends a specific scene by providing a callable function to modify it.
-# The callable recives an instance of the vanilla_scene as the first parameter.
+# The callable receives an instance of the vanilla_scene as the first parameter.
 #
 # Parameters:
 # - scene_vanilla_path (String): The path to the vanilla scene file.
--- a/addons/mod_loader/api/mod.gd
+++ b/addons/mod_loader/api/mod.gd
@@ -74,50 +74,39 @@ static func add_translation(resource_path: String) -> void:
 	ModLoaderLog.info("Added Translation from Resource -> %s" % resource_path, LOG_NAME)
 
 
-# Appends a new node to a modified scene.
+# Refreshes a specific scene by marking it for refresh.
+#
+# This function is useful if a script extension is not automatically applied.
+# This situation can occur when a script is attached to a preloaded scene.
+# If you encounter issues where your script extension is not working as expected,
+# try to identify the scene to which it is attached and use this method to refresh it.
+# This will reload already loaded scenes and apply the script extension.
 #
 # Parameters:
-# - modified_scene (Node): The modified scene where the node will be appended.
-# - node_name (String): (Optional) The name of the new node. Default is an empty string.
-# - node_parent (Node): (Optional) The parent node where the new node will be added. Default is null (direct child of modified_scene).
-# - instance_path (String): (Optional) The path to a scene resource that will be instantiated as the new node.
-#   Default is an empty string resulting in a `Node` instance.
-# - is_visible (bool): (Optional) If true, the new node will be visible. Default is true.
+# - scene_path (String): The path to the scene file to be refreshed.
 #
 # Returns: void
-static func append_node_in_scene(modified_scene: Node, node_name: String = "", node_parent = null, instance_path: String = "", is_visible: bool = true) -> void:
-	var new_node: Node
-	if not instance_path == "":
-		new_node = load(instance_path).instantiate()
-	else:
-		new_node = Node.new()
-	if not node_name == "":
-		new_node.name = node_name
-	if is_visible == false:
-		new_node.visible = false
-	if not node_parent == null:
-		var tmp_node: Node = modified_scene.get_node(node_parent)
-		tmp_node.add_child(new_node)
-		new_node.set_owner(modified_scene)
-	else:
-		modified_scene.add_child(new_node)
-		new_node.set_owner(modified_scene)
+static func refresh_scene(scene_path: String) -> void:
+	if scene_path in ModLoaderStore.scenes_to_refresh:
+		return
+
+	ModLoaderStore.scenes_to_refresh.push_back(scene_path)
+	ModLoaderLog.debug("Added \"%s\" to be refreshed." % scene_path, LOG_NAME)
 
 
-# Saves a modified scene to a file.
+# Extends a specific scene by providing a callable function to modify it.
+# The callable recives an instance of the vanilla_scene as the first parameter.
 #
 # Parameters:
-# - modified_scene (Node): The modified scene instance to be saved.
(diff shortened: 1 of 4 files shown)


addons/mod_loader/internal/scene_extension.gd
```diff
@@ -0,0 +1,43 @@
...
+		ModLoaderLog.debug("Refreshed scene at path: %s" % scene_path, LOG_NAME)
+
+
+# Iterates over the list of scenes to modify and applies the specified edits to each scene.
+static func handle_scene_extensions() -> void:
+	for scene_path in ModLoaderStore.scenes_to_modify.keys():
+		for edit_callable in ModLoaderStore.scenes_to_modify[scene_path]:
+			var cached_scene: PackedScene = load(scene_path)
+			var cached_scene_instance: Node = cached_scene.instantiate()
+			var edited_scene: Node = edit_callable.call(cached_scene_instance)
```
Qubus0, line 25: `var edited_scene: Node` - i wonder if this also works for custom resources. not that we need to account for that here, since we should probably dedicate a method to it

addons/mod_loader/mod_loader.gd
```diff
@@ -180,6 +180,10 @@ func _load_mods() -> void:
 
 	ModLoaderLog.success("DONE: Installed all script extensions", LOG_NAME)
 
+	_ModLoaderSceneExtension.refresh_scenes()
+
+	_ModLoaderSceneExtension.handle_scene_extensions()
+
```
Qubus0, line 186: `ModLoaderLog.success("DONE: Applied all scene extensions", LOG_NAME)`

addons/mod_loader/api/mod.gd
```diff
@@ -171,3 +160,14 @@ static func is_mod_loaded(mod_id: String) -> bool:
...
+
+
+# Deprecated
+# =============================================================================
+
+func append_node_in_scene(modified_scene: Node, node_name: String = "", node_parent = null, instance_path: String = "", is_visible: bool = true) -> void:
+	ModLoaderDeprecated.deprecated_message("ModLoaderMod.append_node_in_scene has been removed, use ModLoaderMod.extend_scene() to edit non preloaded scenes.", "7.0.0")
+
+
+func save_scene(modified_scene: Node, scene_path: String) -> void:
```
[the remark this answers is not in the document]
KANAjetzt: I think we ca actually just delete all deprecation on the 4.x branch, because there where no previews versions used. What do you think?

addons/mod_loader/internal/scene_extension.gd
KANAjetzt: #380

addons/mod_loader/api/mod.gd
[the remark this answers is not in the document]
KANAjetzt: [e1e416d](https://github.com/GodotModding/godot-mod-loader/pull/378/commits/e1e416d9187aecaaa3e9493dd02799d3354d8d9e)

addons/mod_loader/mod_loader.gd
KANAjetzt: [9e0456b](https://github.com/GodotModding/godot-mod-loader/pull/378/commits/9e0456b39b24e68696377fff426b2b248457ca1e)

KANAjetzt:
Should be testable, but first, we need to migrate or rebuild the testing for the 4.x branch 👀

Qubus0 (approved):
new hope for scene modifications 🚀

PR #382, opened 2024-04-04
feat: script extensions sorter checks load order

KANAjetzt:
Cherry-picked #357 

Original description:
> the script extensions sorter will check load order to sort extensions of the same vanilla script
> 
> inheritance is still the first way of sorting, this only sorts conflicting extensions of the same vanilla script

PR #383, opened 2024-04-04
fix: 🐛 fixed missing script parent path

KANAjetzt:
depends on #382


addons/mod_loader/internal/script_extension.gd
```diff
@@ -95,7 +114,7 @@ static func apply_extension(extension_path: String) -> Script:
 	# This is also needed to make Godot instantiate the extended class
 	# when creating singletons.
 	# The actual instance is thrown away.
```
Qubus0, line 116: > The actual instance is thrown away.

no longer accurate since there is no instance

Qubus0 (approved):
only fixes needed are in the other pr

Issue #373, opened 2024-03-21
Self setup missing classes crash


otDan:
We have not updated all the classes that are needed to be added to the project in the self setup process, making any game crash right away if used.

KANAjetzt:
closed by #507

PR #385, opened 2024-04-05
fix: Self setup missing global classes

otDan:
- Adds all the missing classes for the self setup script

Mentioned in https://github.com/GodotModding/godot-mod-loader/issues/373

### DIFF
--- a/addons/mod_loader/mod_loader_setup.gd
+++ b/addons/mod_loader/mod_loader_setup.gd
@@ -19,12 +19,62 @@ const new_global_classes := [
 		"base": "Node",
 		"class": "ModLoaderUtils",
 		"language": "GDScript",
-		"path": "res://addons/mod_loader/mod_loader_utils.gd"
+		"path": "res://addons/mod_loader/internal/mod_loader_utils.gd"
+	}, {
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
are these the same as for godot 3? or more?

otDan:
> are these the same as for godot 3? or more?

not everything is the same I believe which is why I have not pushed it to 3.x yet, will have to check better tomorrow

addons/mod_loader/mod_loader_setup.gd
```diff
@@ -51,7 +101,7 @@ const new_global_classes := [
 		"language": "GDScript",
 		"path": "res://addons/mod_loader/api/mod.gd"
 	}, {
-		"base": "RefCounted",
+		"base": "Reference",
```
KANAjetzt, line 104: Isn't `Reference` 3.x and `RefCounted` 4.x?

otDan: it seems like I was testing with a 3.x game 🤦🏻

otDan:
Actually tested version coming

PR #387, opened 2024-04-06
feat: :sparkles: store the mod source in `ModData`

KANAjetzt:
- Added the `sources` enum to `ModData`. 
- Added the `get_mod_source()` function to  `ModData`. 
- Updated `_init_mod_data()` in `ModLoader` to set the mod source.


Qubus0 (approved):
naming of `set_mod_source` is a little weird, since it both determines the source and then sets it. i'm also not sure if it even needs to be set explicitly or if it could just be part of the setter for `zip_path`. But it's fine like this as well

Qubus0 (approved):

PR #392, opened 2024-04-12
fix: :bug: crash if no scene instance is returned

KANAjetzt:
Fixes a crash where the `scene_instance` is not returned by an `edit_callable`. 

- Added a check for the existence of `edited_scene` before invoking `_save_scene()`. If the `scene_instance` hasn't been returned by the callable, an appropriate error message is logged. 
- Added `SceneEditCallable` subclass to `_ModLoaderSceneExtension` to store the `mod_id` responsible for modifying the scene, enhancing the specificity of the error message.

### DIFF
--- a/addons/mod_loader/api/mod.gd
+++ b/addons/mod_loader/api/mod.gd
@@ -101,7 +101,6 @@ static func refresh_scene(scene_path: String) -> void:
 # Parameters:
 # - scene_vanilla_path (String): The path to the vanilla scene file.
 # - edit_callable (Callable): The callable function to modify the scene.
-# - mod_id (String): The mod_id of your mod, e.g., "namespace-modName".
 #
 # Returns: void
 static func extend_scene(scene_vanilla_path: String, edit_callable: Callable) -> void:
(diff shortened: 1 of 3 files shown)


Qubus0 (commented):
seems a little strange to me that i'd have to pass the mod id, which is then dragged through a new custom class just to us it in an error log at the end.. 
what about an error like this -
```gdscript
"Scene extension of %s failed since the edit callable %s does not return the modified scene_instance" % [scene_path, scene_edit_callable.get_method()]
```
of course get_method won't really be of much help when the method is anonymous, but since this is more of a development error, you should already know where you are doing the wrong thing. and if that is not the case, searching for the scene path will lead you there.

otDan:
> seems a little strange to me that i'd have to pass the mod id, which is then dragged through a new custom class just to us it in an error log at the end.. what about an error like this -
> 
> ```gdscript
> "Scene extension of %s failed since the edit callable %s does not return the modified scene_instance" % [scene_path, scene_edit_callable.get_method()]
> ```
> 
> of course get_method won't really be of much help when the method is anonymous, but since this is more of a development error, you should already know where you are doing the wrong thing. and if that is not the case, searching for the scene path will lead you there.

I don't think it's a bad idea to pass the mod id, it's a bit meh to have but it's good for the future if people actually start using a lot of scene extensions, that way it's easier to even just check who extends what with bigger amounts of mods

Qubus0 (approved):

PR #393, opened 2024-04-12
feat: :sparkles: added `steam_workshop_id` to `ModManifest`

KANAjetzt:
Added an optional field `steam_workshop_id` to `ModManifest`. 

In the editor, it is validated with a basic string length check. Once loaded from a workshop path, it's validated based on the `zip_path`. If invalid, it's overridden by the correct ID retrieved from the `zip_path`.

### DIFF
--- a/addons/mod_loader/resources/mod_manifest.gd
+++ b/addons/mod_loader/resources/mod_manifest.gd
@@ -34,7 +34,7 @@ var tags : PackedStringArray = []
 var config_schema := {}
 var description_rich := ""
 var image: CompressedTexture2D
-
+var steam_workshop_id := ""
 
 # Required keys in a mod's manifest.json file
 const REQUIRED_MANIFEST_KEYS_ROOT = [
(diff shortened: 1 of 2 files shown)


Qubus0 (approved):

addons/mod_loader/resources/mod_manifest.gd
```diff
@@ -458,3 +463,54 @@ static func is_mod_id_valid(original_mod_id: String, check_mod_id: String, type
...
 	return true
+
+
+static func is_string_length_valid(mod_id: String, field: String, string: String, required_length: int, is_silent := false) -> bool:
+	if not string.length() == required_length:
+		if not is_silent:
+			ModLoaderLog.fatal("Invalid length in field \"%s\" of mod \"%s\" it should be \"%s\" but it is \"%s\"." % [field, mod_id, required_length, string.length()], LOG_NAME)
+		return false
+
+	return true
```
otDan, line 474: should this maybe be moved out of `mod_manifest.gd`?

KANAjetzt: I currently don't see a use case outside of the manifest. We can move the validation out if we find one. 👍

otDan (approved):
looks good then

ZackeryRSmith:
This PR seems to introduce a parser error at path.gd:58 as "zip_path_global" is not declared

PR #394, opened 2024-04-18
refactor: :recycle: Changed `disabled_mods` setting to load metadata

KANAjetzt:
- Renamed `disabled_mods` to `deactivated_mods`. 
- Moved default profile creation in `mod_loader.gd` from `_ready()` to `_load_mods()`. This is required so mod configs can be loaded on the first run of the game. 
- Separated loading of mod manifests and mod configs to allow the creation of the default user profile in between. 
- Demoted "No profile" error from error to info because on the first run of the game, the file will not be there. 
- Fixed the `LOG_NAME` of `_ModLoaderPath`. 
- Changed the `MOD_CONFIG_DIR_PATH` from `"user://configs"` to `"user://mod_configs"` to make it clearer what kind of configs are stored. 
- Fixed an issue where `current_config` was `null` because it wasn't set in `ModData._load_config`.

### DIFF
--- a/addons/mod_loader/resources/options_profile.gd
+++ b/addons/mod_loader/resources/options_profile.gd
@@ -5,7 +5,10 @@ extends Resource
 @export var enable_mods: bool = true
 @export var locked_mods: Array[String] = []
 @export var log_level := ModLoaderLog.VERBOSITY_LEVEL.DEBUG
-@export var disabled_mods: Array[String] = []
+## Mods in this array will only load metadata but will not apply any modifications.
+## This can be used to deactivate example mods.
+## Mods in this array are overridden by the settings in the user profile.
+@export var deactivated_mods: Array[String] = []
 @export var allow_modloader_autoloads_anywhere: bool = false
 @export var steam_id: int = 0
 @export_dir var override_path_to_mods = ""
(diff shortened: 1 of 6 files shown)


addons/mod_loader/resources/options_profile.gd
```diff
@@ -8,7 +8,10 @@ extends Resource
 @export var enable_mods: bool = true
 @export var locked_mods: Array[String] = []
 @export var log_level := ModLoaderLog.VERBOSITY_LEVEL.DEBUG
-@export var disabled_mods: Array[String] = []
+## Mods in this array will only load metadata but will not apply any modifications.
+## This can be used to deactivate example mods.
+## Mods in this array are overridden by the settings in the user profile.
+@export var deactivated_mods: Array[String] = []
```
otDan, line 14: why are we changing it to deactivated? seems like a pointless breaking change?

Qubus0: deactivated makes more sense to me, so i think the rename is fine. but yes, this is breaking and no we can't overlook that anymore since we have windowkill now which relies on it. so we need a proper deprecation both for that and for mod_configs

KANAjetzt:
I have moved the changes into separate issues so we can tackle them once we have completed the major changes for v7.

- [ ] #464 
- [ ] #465
- [ ] #466
- [ ] #468

PR #398, opened 2024-04-22
feat: :sparkles: added `is_mod_active` to `ModLoaderMod`

### LINKED ISSUE #355: add a nice way for mods to check if another mod is in use (loaded and active)
Currently something like this needs to be used to check if a mod is in use. 

```gdscript
if ModLoaderMod.is_mod_loaded(mod_id) and ModLoaderMod.get_mod_data(mod_id).is_active:
```

Since this is likely a common check to mesh well with other mods, we might want to add a new ModLoaderMod method to do it in one go.

only using `is_mod_loaded` would have been the old way to check this, but this does not work anymore since we are loading all mods to access the manifest data

KANAjetzt:
Added  `static func is_mod_active(mod_id: String) -> bool:` to `ModLoaderMod`.

closes #355 (maybe we want to keep it open for 3.5?)

### DIFF
--- a/addons/mod_loader/api/mod.gd
+++ b/addons/mod_loader/api/mod.gd
@@ -161,3 +161,17 @@ static func is_mod_loaded(mod_id: String) -> bool:
 		return false
 
 	return true
+
+
+# Returns true if the mod with the given mod_id was successfully loaded and is currently active.
+#
+# Parameters:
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
can technically be shortened to a return, but i also like being explicit sometimes

PR #399, opened 2024-04-23
refactor: :fire: removed `steam_workshop_enabled` option

### LINKED ISSUE #384: Delete `steam_workshop_enabled` from options
With PR #369, we added the option to load from multiple sources and deprecated the `steam_workshop_enabled` option. In PR #381, we removed all deprecations, but the option wasn't deleted.

KANAjetzt:
It has been replaced by the `load_from_steam_workshop` option in #369 

closes #384


KANAjetzt:
> windowkill uses load_from_steam_workshop already, right?
yes 👍

PR #402, opened 2024-06-04
feat: ✨ `create_deactivated_mods_profile` option added 

KANAjetzt:
Added `create_deactivated_mods_profile` and `deactivated_mods_profile_name` option. This can be used after a major game update to prevent crashes.

Tested with Windowkill 👍 

> [!IMPORTANT]  
> depends on #394


KANAjetzt:
I'm closing this. It is a bit stale, and enough has changed since then that we need to take a fresh look at it.

PR #404, opened 2024-07-01
fix: do not ignore manual installers

ZackeryRSmith:
This is labeled as a fix, why? Better yet why make this PR in the first place? Isn't this feature here for good reason?

I do like the idea of ignoring the files not required by AssetLib downloads, but doing it this way has the caveat of ignoring these files when downloading the repo (at least from Github directly). To me is a **big** no no. It restricts the people who actually want to use the self setup, with the benefit of removing a couple extra files from the AssetLib download.

Overall I think these files are small and quite harmless to AssetLib downloads, while it's harmful to make it harder for people who need the self setup to get it.

> [!NOTE]  
> This PR targets the 4.x branch but applies to 3.x too


ZackeryRSmith:
Seems extra commits have been dragged along here... Let me fix that.

KANAjetzt:
We can provide the full self setup version via the [releases](https://github.com/GodotModding/godot-mod-loader/releases) page.

ZackeryRSmith:
> We can provide the full self setup version via the [releases](https://github.com/GodotModding/godot-mod-loader/releases) page.

Well as it stands it isn't included, but if any changes need to get made to the self setup another release would have to be made. Just for one file change.

PR #413, opened 2024-09-16
docs: update wiki URLs in code

ZackeryRSmith:
As noticed by @Qubus0 wiki URLs in the code needed to be changed to use the GMLWiki


KANAjetzt (approved):
thanks 👍

PR #414, opened 2024-10-14
fix: broken TranslationServer when added translation fails to load

pirey0:
Adds a null check to ensure broken translation files don't break the game's localization.

Godots translation.cpp has no safety checks when calling TranslationServer::add_translation. (as of 4.2.2)
If for any reason a null-object is passed to it it will cause the translation server to break for the entire session.


KANAjetzt (approved):
Thanks 👍

PR #416, opened 2024-10-16
Feat zip peeking

### LINKED ISSUE #407: Loading prevention for incompatible game version
It would be good to consider a method of adding prevention from loading a mod in case of incompability.

There is currently arguments in json manifest schema

```json
	  "compatible_mod_loader_version": "4.0.0",
	  "compatible_game_version": ["4.0.3"],
```
while there is a method for handling mod_loader_version `_handle_compatible_mod_loader_version` there is non for game_version nor game engine

`compatible_game_version` if it's minor changes probably should be implemented by dev directly although there could be added helper method.

The real problem if devs would decide to change engine version as due to syntax changes it would crash the game while trying to load files. So would propose adding `compatible_engine_version`, where engine version could be easily retrieve by `Engine.get_version_info()`

Qubus0:
big pr with some suboptimal mixing of things to get them fixed for dome, sorry about those. 

the core of the zip peeking pr is about having info about mod metadata before the mod is loaded by load_resource_pack. this lets us compare the game version (and other data, in following prs) from the zip with the one set by the game. if a version mismatch is detected the mod won't be loaded, which will reduce crashes.

this basically flips the order of operations around from previously 
1. loading resource packs
2. parsing the metadata 
to
1. reading metadata from the zip or filesystem
2. loading resource packs only if the metadata indicates they can be loaded

***

optimally this would be the new process, but i had to take a few shortcuts to quickly get it ready for the dome patch
- [x] get all the mod paths
- [ ] validate all required files exist in the zip/unpacked folder
- [x] load text data from the paths
- [x] init the mod data from the collected file paths
- [x] read manifest and add to mod data
- [ ] read config and add to mod data
- [ ] check the mod profile
- [ ] filter and collect disable reasons
	- [ ] not loadable due to errors
	- [ ] mod enabled/disabled in profile
	- [x] wrong game version -> if lower.. major: off, minor: log warn, patch: log info
	- [ ] wrong mod loader version -> same logic
- [x] figure out which files get hooks
- [x] dynamically generate pck only for those files
- [x] apply pck
- [x] load mods that don't have errors
- [x] apply extensions and so on

shortcut way
- [x] get all the mod paths
- [x] load text data from the paths
- [x] minimally init the mod data
- [x] read manifest and add to mod data
- [x] filter and collect errors/disable reasons (only version and bad manifest)
- [x] load mods that don't have errors
- [x] apply extensions and so on

## Other notes for the future

- mod data will always be available
   - so all mod state should be stored in it, it becomes the source of truth
- deprecate Store ml_options.disabled_mods and locked_mods
- i lost some logging when reading mod metadata
- profile _set_mod_state can force to enable a mod even with load warnings
- kinda disabled loading unpacked mods outside the editor
- had to disable reload_mods


addons/mod_loader/api/profile.gd
```diff
@@ -25,6 +25,17 @@ static func enable_mod(mod_id: String, user_profile:= ModLoaderStore.current_use
 	return _set_mod_state(mod_id, user_profile.name, true)
 
 
+## Forces enabling a mod - it will be loaded on the next game start even if load warnigns are present[br]
+## [br]
+## [b]Parameters:[/b][br]
+## - [code]mod_id[/code] ([String]): The ID of the mod to enable.[br]
+## - [code]user_profile[/code] ([ModUserProfile]): (Optional) The user profile to enable the mod for. Default is the current user profile.[br]
+## [br]
+## [b]Returns:[/b] [bool]
```
KANAjetzt:
```suggestion
## Forces a mod to enable, ensuring it loads at the next game start, regardless of load warnings.[br]
## [br]
## [b]Parameters:[/b][br]
## - [code]mod_id[/code] ([String]): The ID of the mod to enable.[br]
## - [code]user_profile[/code] ([ModUserProfile]): (Optional) The user profile for which the mod will be enabled. Defaults to the current user profile.[br]
## [br]
## [b]Returns:[/b] [bool]: True if the mod is successfully set to enable, False otherwise.
```

addons/mod_loader/internal/file.gd
```diff
@@ -45,102 +44,87 @@ static func _get_json_string_as_dict(string: String) -> Dictionary:
...
+		func(file_name: String):
+			return file_name.get_extension() == "zip"
+	).map(
+		func(file_name: String):
+			var global_path := ProjectSettings.globalize_path(folder_path.path_join(file_name))
+			ModLoaderLog.debug("Found mod ZIP: %s" % global_path, LOG_NAME)
+			return global_path
+	)
+
+	# only assign casts to the nested Array type we want to return
```
KANAjetzt: I don't understand that comment.
- What is meant with casts
- Why only assign casts to the nested Array?
- What nested Array?

addons/mod_loader/internal/file.gd
```diff
@@ -209,17 +193,34 @@ static func remove_file(file_path: String) -> bool:
...
+		return zip_file_exists(path, zip_path)
+
 	return FileAccess.file_exists(path)
 
 
 static func dir_exists(path: String) -> bool:
 	return DirAccess.dir_exists_absolute(path)
 
 
+static func zip_file_exists(path: String, zip_path: String = "") -> bool:
```
KANAjetzt: This checks if a file exists inside a zip, right? I think something like this would make it clearer.

```suggestion
static func file_in_zip_exists(path: String, zip_path: String = "") -> bool:
```

addons/mod_loader/internal/hooks.gd
```diff
@@ -0,0 +1,29 @@
+class_name _ModLoaderHooks
+extends Object
+
+const LOG_NAME := "ModLoader:Hooks"
```
KANAjetzt:
```suggestion
class_name _ModLoaderHooks
extends Object


# This Class provides util functions for working with Mod Hooks.
# Currently all of the included functions are internal and should only be used by the mod loader itself.

const LOG_NAME := "ModLoader:Hooks"
```

addons/mod_loader/internal/mod_hook_packer.gd
```diff
@@ -9,6 +9,7 @@ const LOG_NAME := "ModLoader:ModHookPacker"
 
 
 static func start() -> void:
+	ModLoaderLog.info("Generating mod hooks .pck", LOG_NAME)
```
KANAjetzt:
```suggestion
	ModLoaderLog.info("Generating mod hooks .zip", LOG_NAME)
```

addons/mod_loader/mod_loader.gd
```diff
@@ -69,8 +69,20 @@ func _init() -> void:
...
+		var path := _ModLoaderPath.get_unpacked_mods_dir_path()
+		for mod_dir in DirAccess.get_directories_at(path):
+			var manifest_path := path.path_join(mod_dir).path_join("manifest.json")
+			var manifest_dict := _ModLoaderFile.get_json_as_dict(manifest_path)
+			_load_metadata(manifest_dict)
+
+	var mod_zip_paths := _get_mod_zip_paths()
+	for zip_path in mod_zip_paths:
+		var manifest_dict := _ModLoaderFile.get_json_as_dict_from_zip(zip_path, "manifest.json")
+		_load_metadata(manifest_dict, zip_path)
```
KANAjetzt, line 82: I feel like there should be a load manifest function in `_ModLoaderFile` ?

addons/mod_loader/mod_loader.gd
```diff
@@ -251,101 +264,79 @@ func _check_autoload_positions() -> void:
...
+		# no longer exist in the file system. So this warning basically says
+		# "don't use ZIPs with unpacked mods!"
+		# https://github.com/godotengine/godot/issues/19815
+		# https://github.com/godotengine/godot/issues/16798
+		if OS.has_feature("editor") and not ModLoaderStore.has_shown_editor_zips_warning:
+			ModLoaderLog.warning(str(
+				"Loading any resource packs (.zip/.pck) with `load_resource_pack` will WIPE the entire virtual res:// directory. ",
+				"If you have any unpacked mods in ", _ModLoaderPath.get_unpacked_mods_dir_path(), ", they will not be loaded. ",
+				"Please unpack your mod ZIPs instead, and add them to ", _ModLoaderPath.get_unpacked_mods_dir_path()), LOG_NAME)
+			ModLoaderStore.has_shown_editor_zips_warning = true
```
KANAjetzt, line 331: Does `only_once` not work in this case?
https://github.com/GodotModding/godot-mod-loader/blob/d7a28c31902cc0e08a8d756b23b69357160709ee/addons/mod_loader/api/log.gd#L135

addons/mod_loader/mod_loader_store.gd
```diff
@@ -138,6 +138,8 @@ var ml_options := {
 
 	# Application's Steam ID, used if workshop is enabled
 	steam_id = 0,
+	# Application's version following semver
+	semantic_version = "0.0.0",
```
KANAjetzt, line 142: what if a game uses non semantic version numbers?

KANAjetzt (approved):
Massive rework - thanks a lot! 🩵 This will require intensive testing, but I’m sure with this implemented in Dome Keeper, we will find all the issues 😄

I’ll start testing this with my projects. Currently, I don’t see anything holding us back from merging it 👍

addons/mod_loader/internal/file.gd
Qubus0: the method is supposed to return `Array[String]`, the nested array type. `Array.filter` and `map` only return raw `Array`, which causes a type error when returned directly if the nested return type is set. godot can't type cast to nested array types with `as` sadly, so we can't just `return files as Array[String]`. The only way to change the type is to put all the elements into an array that already has the nested type, which can be done with `.assign()`

addons/mod_loader/internal/file.gd
Qubus0: good shout, `file_exists_in_zip` probably makes the most sense

addons/mod_loader/mod_loader.gd
Qubus0: yeah that could be cleaner, added todo

addons/mod_loader/mod_loader.gd
Qubus0: it should, yeah. i didn't change anything here, just moved it.

addons/mod_loader/mod_loader_store.gd
Qubus0: then they can either leave it like this, which essentially disables version checking, implement their own logic since we can't preemtively cover all strange version standards, or switch to semver, which is probably the easiest. it's the main version standard for a reason

Qubus0 (commented):
followup pr todo
- [ ] move manifest file loading to file.gd
- [ ] fix reload mods
- [x] doc comments for hooks
- [ ] use log `only_once` instead of `ModLoaderStore.has_shown_editor_zips_warning`

addons/mod_loader/internal/file.gd
KANAjetzt: Oh, I completely read that wrong because I didn't connect assign with `.assign()` 😄 
Might be worth to extend the comment a bit with what you added here.

addons/mod_loader/mod_loader_store.gd
KANAjetzt: Maybe we could add an option to enable or disable this, or even allow the developer to implement their own validation. Just a thought for the future.

addons/mod_loader/internal/file.gd
```diff
@@ -45,102 +44,87 @@ static func _get_json_string_as_dict(string: String) -> Dictionary:
...
+	var zip_paths: Array[String] = []
+
+	var files := Array(DirAccess.get_files_at(folder_path))\
+	.filter(
+		func(file_name: String):
+			return file_name.get_extension() == "zip"
+	).map(
+		func(file_name: String):
+			var global_path := ProjectSettings.globalize_path(folder_path.path_join(file_name))
+			ModLoaderLog.debug("Found mod ZIP: %s" % global_path, LOG_NAME)
```
pirey0: Nitpick: 
I don't think there should be sideffects in a functional call. 
In this case debug() has no real way of failing so it doesn't matter much, but for me the logging call should happen outside of this function alltogether, or at least after the mapping.

addons/mod_loader/internal/file.gd
```diff
@@ -209,17 +193,34 @@ static func remove_file(file_path: String) -> bool:
...
 	return FileAccess.file_exists(path)
 
 
 static func dir_exists(path: String) -> bool:
 	return DirAccess.dir_exists_absolute(path)
 
 
+static func file_exists_in_zip(path: String, zip_path: String = "") -> bool:
+	var reader := zip_reader_open(zip_path)
+	return reader.file_exists(path.trim_prefix("res://"))
```
pirey0, line 210: reader can be undefined (see below). I would handle it gracefully and return false.
I'm unsure of what would happen right now in that case in a release build.

addons/mod_loader/internal/path.gd
```diff
@@ -5,25 +5,31 @@ extends RefCounted
...
 # folder, ie "res://mods" or the OS's equivalent, as well as the configs path
 static func get_local_folder_dir(subfolder: String = "") -> String:
+	return get_game_install_dir().path_join(subfolder)
+
+
+static func get_game_install_dir() -> String:
 	var game_install_directory := OS.get_executable_path().get_base_dir()
 
-	if OS.get_name() == "OSX":
+	if OS.get_name() == "macOS":
```
pirey0, line 21: from what i can see in the c++ code the feature flag is "macos" (no capitalization), is it working correctly with the capitalized OS?
image
image

addons/mod_loader/internal/third_party/steam.gd
```diff
@@ -6,46 +6,45 @@ const LOG_NAME := "ModLoader:ThirdParty:Steam"
...
 
 	ModLoaderLog.info("Checking workshop items, with path: \"%s\"" % workshop_folder_path, LOG_NAME)
 
 	var workshop_dir := DirAccess.open(workshop_folder_path)
 	if workshop_dir == null:
-		ModLoaderLog.error("Can't open workshop folder %s (Error: %s)" % [workshop_folder_path, DirAccess.get_open_error()], LOG_NAME)
-		return {}
+		ModLoaderLog.error("Can't open workshop folder %s (Error: %s)" % [workshop_folder_path, error_string(DirAccess.get_open_error())], LOG_NAME)
+		return []
 	var workshop_dir_listdir_error := workshop_dir.list_dir_begin() # TODOGODOT4 fill missing arguments https://github.com/godotengine/godot/pull/40547
```
pirey0, line 23: Nitpick: `workshop_dir_listdir_error ` could just be `result` or `error`

pirey0 (commented):
Are we sure about `"macOS"` working correctly over `"macos"`?

addons/mod_loader/internal/file.gd
Qubus0: yep absolutely makes sense

addons/mod_loader/internal/file.gd
Qubus0: good catch

addons/mod_loader/internal/path.gd
Qubus0: yep, that's get_name() though, not the flag. confused me too. we should probably decide on a single one - i think it was originally chosen because has_feature [was the same with web and without](https://docs.godotengine.org/en/3.5/tutorials/export/feature_tags.html), which [has changed since](https://docs.godotengine.org/en/stable/tutorials/export/feature_tags.html)

Qubus0 (commented):
no new todos i'll fix those right away

PR #408, opened 2024-08-22
feat: ✨ Mod Hooks

KANAjetzt:
# Modding Hooks

- Export Plugin to create mod hooks
- New dynamic creation of a .zip file including only necessary mod hooks
- New Modding API to modify game scripts

## New Class `_ModLoaderModHookPreProcessor`

This class is used to process the source code from a script at a given path.

- Replaces all vanilla function names with something like `vanilla_2924080498__ready()`
- Replaces all instances of `super()` with `super.vanilla_method_name()`
- Check if the function is a getter or setter
- Check if a function returns something and potential return types
- Get function parameters with and without types
- Handle function parameter defaults properly
- Check for inner classes
- Check if static or not
- Check if a function is marked as moddable or not moddalbe
- Create an imposter function with all the information collected
  - This function includes calls to the mod Callables added via `ModLoaderMod.add_hook()`

### How to use:

- Instance a new `_ModLoaderModHookPreProcessor`
- Call `process_script(path)` to start processing the source code of a script
- `hook_pre_processor.process_begin()` can be called to clear the hashmap storing the hook data.

### Result

<details>
<summary>
Before processing
</summary>

```gdscript
extends LevelBase

@onready var game_times: Node3D = %GameTimes

func _ready() -> void:
	super()
	Global.current_arena_index = level_data.level_id

	Global.game_won.connect(_on_game_won)
	Global.player.camera.tween_completed.connect(_on_player_camera_tween_completed)

func _on_portal_hit_detected(destination: LevelData) -> void:
	Global.player.is_input_disabled = true
	Global.blend()
	await get_tree().create_timer(0.1).timeout
	Global.player.teleport(destination.start_transform)
	Global.player.activate_particles()
	await get_tree().create_timer(0.1).timeout
	destination.ref.activate_camera()

func _on_game_won(first: bool) -> void:
	%PortalNewGame.deactivated = false
	if first and not Global.settings.gameplay_ui_show_level_timer:
		Global.settings.gameplay_ui_show_level_timer = true
	if first and not Global.settings.gameplay_ui_show_overall_timer:
		Global.settings.gameplay_ui_show_overall_timer = true

	game_times.get_child(-1).text = "%s | %s" % [game_times.get_child(-1).text, Global.score]
	add_time(Global.time_game)

func add_time(time: int) -> void:
	var new_label := Label3D.new()
	new_label.text = Global.hud.format_stopwatch(time)
	new_label.font_size = 900
[...]

PR #420, opened 2024-11-12
fix: Use counter instead of stack in get_closing_paren_index()

Qubus0:
fixes todo from #408

- [x] [Use counter instead of stack in get_closing_paren_index()](https://github.com/GodotModding/godot-mod-loader/pull/408/files#r1832995998)


KANAjetzt (approved):
👍

PR #425, opened 2024-11-13
fix: :bug: fixed super fixing

KANAjetzt:
With the changes from #420, we need to start the search for `closing_paren_index` at `func_def_end - 1`.

But this needs a deeper look. based on my debugging, it seems there’s more wrong with how we’re processing that part.

<details><summary>Some print debugging</summary>
<p>

```
  text on func_def_end: ) ->
  method_name: _ready
  closing_paren_index: -1
  func_def_end: 83
  ---
  extends Portal
  
  
  @export var label_level_name_override := "New Game"
  
  
  func _ready() -> void:
      super()
      label_level_name.text = label_level_name_override
  
  
  func _on_hit_detector_body_entered(body: Node3D) -> void:
      super(body)
      Global.reset_game()
  
  No matching closing parenthesis
  text on func_def_end: body
  method_name: _on_hit_detector_body_entered
  closing_paren_index: -1
  func_def_end: 210
  ---
  extends Portal
  
  
  @export var label_level_name_override := "New Game"
  
  
  func vanilla_4007268940__ready() -> void:
      super()
      label_level_name.text = label_level_name_override
  
  
  func _on_hit_detector_body_entered(body: Node3D) -> void:
      super(body)
      Global.reset_game()
  
  No matching closing parenthesis
```

- I think `text on func_def_end` should always start with `)`. It looks like the result is shifted for the second method, and if this pattern continues, things will start to break again.

</p>
</details> 

For now, this should be enough to keep things moving.

### DIFF
--- a/addons/mod_loader/internal/mod_hook_preprocessor.gd
+++ b/addons/mod_loader/internal/mod_hook_preprocessor.gd
@@ -152,7 +152,9 @@ static func get_function_parameters(method_name: String, text: String, is_static
 	if not is_top_level_func(text, result.get_start(), is_static):
 		return get_function_parameters(method_name, text, is_static, result.get_end())
 
-	var closing_paren_index := get_closing_paren_index(opening_paren_index, text)
+	# Shift the func_def_end index back by one to start on the opening parentheses.
+	# Because the match_func_with_whitespace().get_end() is the index after the opening parentheses.
+	var closing_paren_index := get_closing_paren_index(opening_paren_index - 1, text)
 	if closing_paren_index == -1:
 		return ""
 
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
`match_func_with_whitespace` actually matches until including the first opening `(` of a function, `func _ready(` for example. the following characters are "shifted" because they don't matter, the end of the full function declaration can only be after `get_closing_paren_index` finds the correct closing index. 
since `match_func_with_whitespace` always includes the (, we actually always have to subtract 1, since `get_closing_paren_index` starts from the exact index where that ( is, not one after it. 
this code is actually more stable now, the previous version would have caused more issues eventually.
we should document these better though, and better naming is also in order

KANAjetzt:
Oh right - in the morning everything seems clearer 😄
So I change it in `get_closing_paren_index()` then.

```gdscript
static func get_closing_paren_index(opening_paren_index: int, text: String) -> int:
	# Use a stack counter to match parentheses
	var stack := 0
	var closing_paren_index := opening_paren_index - 1 # Shift the index back by one to start before the opening parentheses.
        [...]
```

Qubus0:
nono, where it is is completely correct. that is a thing specific to using `get_closing_paren_index` right after `match_func_with_whitespace`. though that's the only way we use it currently since we do `-1` in both `fix_method_super` and `get_function_parameters`

KANAjetzt:
ok got it now 😄 then I add the comment where we call `get_closing_paren_index()` in `fix_method_super()` and `get_function_parameters()`

PR #427, opened 2024-11-14
feat: :sparkles: add `is_silent` to `process_script()`

KANAjetzt:
This allows the Dev Tool to process scripts without errors.
- https://github.com/GodotModding/godot-mod-tool/pull/111#issuecomment-2475015380

### DIFF
--- a/addons/mod_loader/internal/mod_hook_preprocessor.gd
+++ b/addons/mod_loader/internal/mod_hook_preprocessor.gd
@@ -36,9 +36,10 @@ func process_begin() -> void:
 	hashmap.clear()
 
 
-func process_script(path: String) -> String:
+func process_script(path: String, is_silent := false) -> String:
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
not the prettiest but does work

Qubus0:
hm wait i had another quick think, maybe we can just add a verbose version to call alternatively. should also technically help performance a miniscule amount

```gdscript
func process_script_verbose(path: String) -> String:
	var start_time := Time.get_ticks_msec()
	ModLoaderLog.debug("Start processing script at path: %s" % path, LOG_NAME)
	var processed := process_script(path)
	ModLoaderLog.debug("Finished processing script at path: %s in %s ms" % [path, Time.get_ticks_msec() - start_time], LOG_NAME)
	return processed
```

KANAjetzt:
closed in favor of 
- #430

PR #428, opened 2024-11-14
style: :fire: removed `%` from hook template string

KANAjetzt:
Removed the `%` in the hook function template to improve readability.


Qubus0 (approved):
✨

PR #429, opened 2024-11-14
 feat: disable hook check when not needed

Qubus0:
resolves todo from #408 

- [x] [Find a way to get rid of the ModLoaderStore.get("any_mod_hooked") check, it invalidates the performance benefit of ModLoaderStore.any_mod_hooked.](https://github.com/GodotModding/godot-mod-loader/pull/408/files#r1832930546)
    - [x] Skip this check altogether if hooks are dynamically generated


addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -68,7 +68,7 @@ func process_script(path: String) -> String:
 		var hash_before := _ModLoaderHooks.get_hook_hash(path, method.name, true)
 		var hash_after := _ModLoaderHooks.get_hook_hash(path, method.name, false)
 		var hash_before_data := [path, method.name,true]
```
KANAjetzt:
```suggestion
		var hash_before_data := [path, method.name, true]
```

KANAjetzt (approved):
I'm sure you tested this 👍

PR #430, opened 2024-11-14
feat: split off verbose hook preprocessing

Qubus0:
depends on #429 

slightly improved alternative to #427


KANAjetzt (commented):
Also verbose in `_ModLoaderModHookPacker`?

KANAjetzt (approved):

PR #431, opened 2024-11-14
fix: proper logging of empty Array()

Qubus0:
apparently the implicit conversion of an empty array created with `Array()` breaks (at least in g4.2) - explicitly converting it with str() fixes it though.

### DIFF
--- a/addons/mod_loader/internal/file.gd
+++ b/addons/mod_loader/internal/file.gd
@@ -119,7 +119,7 @@ static func get_zip_paths_in(folder_path: String) -> Array[String]:
 		func(file_name: String):
 			return ProjectSettings.globalize_path(folder_path.path_join(file_name))
 	)
-	ModLoaderLog.debug("Found mod ZIPs: %s" % files, LOG_NAME)
+	ModLoaderLog.debug("Found mod %s ZIPs: %s" % [files.size(), str(files)], LOG_NAME)
 
 	# only .assign()ing to a typed array lets us return Array[String] instead of just Array
 	zip_paths.assign(files)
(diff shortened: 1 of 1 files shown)


addons/mod_loader/internal/file.gd
```diff
@@ -119,7 +119,7 @@ static func get_zip_paths_in(folder_path: String) -> Array[String]:
 		func(file_name: String):
 			return ProjectSettings.globalize_path(folder_path.path_join(file_name))
 	)
-	ModLoaderLog.debug("Found mod ZIPs: %s" % files, LOG_NAME)
+	ModLoaderLog.debug("Found mod %s ZIPs: %s" % [files.size(), str(files)], LOG_NAME)
```
KANAjetzt: How does this look printed out? 
Bit confused why we got two %s now.

Qubus0: it logs the count of found zips too now to avoid confusion when 0 zips are found and it still says it found some but shows an empty array. now it is
`Found mod 0 ZIPs: []` or
`Found mod 2 ZIPs: ["bla", "bla2"]`

KANAjetzt: Can we do `Found 2 mod ZIPs: ["bla", "bla2"]`?

```suggestion
	ModLoaderLog.debug("Found %s mod ZIPs: %s" % [files.size(), str(files)], LOG_NAME)
```

Qubus0: this is the first time i read that i put the %s one word too late i swear it was always before the "mod"

KANAjetzt (approved):
approved 🎉

PR #432, opened 2024-11-14
test: add new gut test cases for the hook preprocessor

Qubus0:
just test cases for a few methods that should not break


KANAjetzt (approved):
Nice - We don't have a test setup for 4.x until now. Maybe we want to look at other options than GUT for Godot 4?

Qubus0:
we could, yeah. just used what we had and i'm not bothered by it yet. could use a bigger thought but maybe we're also overthinking a solution where good enough is good enough

PR #434, opened 2024-11-15
test: :test_tube: Ported tests to 4.x

KANAjetzt:
Ported the test workflow from the 3.x branch  

## Notable Changes  
- Updated the GUT version to the [latest release v9.3.0](https://github.com/bitwes/Gut/releases/tag/v9.3.0)  
- Updated the Godot version:  
  - Currently running tests only on 4.3  
  - No separate headless version, now using `--headless` instead  
  - Global class data is no longer stored in the project file, requiring a `--import` run before tests are executed  

## To-Do  
- [ ] Run tests on additional Godot versions:  
  - [ ] 4.2.2  
  - [ ] 4.1.4  
- [x] Resolve two currently failing tests:  
  - [x] `test_match_func_with_whitespace`  
  - [x] `test_handle_script_extensions`

### DIFF
--- a/addons/mod_loader/options/profiles/default.tres
+++ b/addons/mod_loader/options/profiles/default.tres
@@ -1,14 +1,24 @@
-[gd_resource type="Resource" load_steps=2 format=2]
+[gd_resource type="Resource" script_class="ModLoaderOptionsProfile" load_steps=2 format=3 uid="uid://du84b0sewwguy"]
 
-[ext_resource path="res://addons/mod_loader/resources/options_profile.gd" type="Script" id=1]
-
+[ext_resource type="Script" path="res://addons/mod_loader/resources/options_profile.gd" id="1"]
 
 [resource]
-script = ExtResource( 1 )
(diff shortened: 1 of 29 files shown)


addons/mod_loader/options/profiles/default.tres
Qubus0, line 1: i think we don't want to commit the automatic changes godot makes to the profiles since they seem to break 4.1, we'll have to think of a better solution here

Qubus0 (approved):
i think we can clean up some of the "expected x was y instead" extra messages we added since that is logged anyway.

otherwise nice! thank you

addons/mod_loader/options/profiles/default.tres
KANAjetzt: addons/mod_loader should be untouched by this PR anyway 👀 But good point.

_Done in [9835a27](https://github.com/GodotModding/godot-mod-loader/pull/434/commits/9835a27e88c6d320ed10e0dd2f541611b67bd958)_

Issue #426, opened 2024-11-13
dynamic hooking: selectively only hook targeted methods instead of whole file


Qubus0:
by hooking only what's necessary we should be able to work out the last bit of performance and also have less risk for potential side effects

KANAjetzt:
closed by [#534](https://github.com/GodotModding/godot-mod-loader/pull/534)

PR #436, opened 2024-11-16
feat: ✨ param and return manipulation for mod hooks

Qubus0:
allows modders to manipulate parameters and return values similarly to how script extensions work.
also docs

test project with unpacked mod: [project.zip](https://github.com/user-attachments/files/17791098/project.zip)


addons/mod_loader/api/hook.gd
```diff
@@ -0,0 +1,47 @@
...
+## For examples, see [method ModLoaderMod.add_hook].
+
+
+## The reference object is usually the [Node] that the vanilla script is attached to. [br]
+## If the hooked method is [code]static[/code], it will contain the [GDScript] itself.
+var reference_object: Object
+
+var _return_val: Variant
+# The next linkage in the chain
+var _next_linkage: ModLoaderHook
```
KANAjetzt: I personally would prefer `_next_link`, `next_hook` or `_next_mod_hook_call`.
I never heard linkage before 😄

addons/mod_loader/api/hook.gd
```diff
@@ -0,0 +1,47 @@
...
+extends RefCounted
+## Small class to pass data between mod hook calls.[br]
+## For examples, see [method ModLoaderMod.add_hook].
+
+
+## The reference object is usually the [Node] that the vanilla script is attached to. [br]
+## If the hooked method is [code]static[/code], it will contain the [GDScript] itself.
+var reference_object: Object
+
+var _return_val: Variant
```
KANAjetzt: I think if we do `reference_object` we can also do `_return_value`

addons/mod_loader/api/mod.gd
```diff
@@ -41,10 +46,89 @@ static func install_script_extension(child_script_path: String) -> void:
...
+## ## then we can change any variables it has
+## func change_version(hook: ModLoaderHook) -> void:
+##     # Using a typecast here (with "as") can help with autocomplete and avoiding errors
+##     var main_node := hook.reference_object as MainGame
+##     main_node.version = "Modloader Hooked!"
+##     # _ready, which we are hooking, does not have any arguments
+##     hook.execute_next()
+##
+##
+## ## Parameters can be manipulated easily by changing what is passed into .execute_next()
```
KANAjetzt: Nitpick: No need for doc comment above function names in the example?

addons/mod_loader/internal/hooks.gd
```diff
@@ -7,27 +7,36 @@ extends Object
...
-static func call_hooks(self_object: Object, args: Array, hook_hash:int) -> void:
+static func call_hooks(vanilla_method: Callable, args: Array, hook_hash: int) -> Variant:
 	var hooks = ModLoaderStore.modding_hooks.get(hook_hash, null)
-	if hooks:
-		for mod_func: Callable in hooks:
-			mod_func.bind(self_object).callv(args)
+	if not hooks:
+		return vanilla_method.callv(args)
+
+	# Create a hook chain which will recursively call down until the vanilla method is reached
```
KANAjetzt: I would love a small example and a more detailed description on how the chain works here.

KANAjetzt (approved):
Tested with the provided example project, in editor and as exported game.

Amazing work - thanks!

addons/mod_loader/api/hook.gd
Qubus0: ima delete it instead 😤
finally worked out how to not need it, not sure why there was a knot in my brain that didn't let me before

addons/mod_loader/internal/hooks.gd
Qubus0: added better explanations in `hook.gd`

addons/mod_loader/internal/hooks.gd
```diff
@@ -7,29 +7,39 @@ extends Object
...
 	ModLoaderStore.any_mod_hooked = true
-	var hash = get_hook_hash(script_path,method_name,is_before)
+	var hash = get_hook_hash(script_path, method_name)
+
 	if not ModLoaderStore.modding_hooks.has(hash):
 		ModLoaderStore.modding_hooks[hash] = []
 	ModLoaderStore.modding_hooks[hash].push_back(mod_callable)
-	ModLoaderLog.debug("Added hook script: \"%s\" %s method: \"%s\""
-		% [script_path, "before" if is_before else "after", method_name ], LOG_NAME
+	ModLoaderLog.debug('Added hook "%s" to to method: "%s" in script: "%s"'
```
pirey0: duplicate "to" `to to`

addons/mod_loader/internal/hooks.gd
```diff
@@ -7,29 +7,39 @@ extends Object
...
-	var hooks = ModLoaderStore.modding_hooks.get(hook_hash, null)
-	if not hooks:
-		return
+static func call_hooks(vanilla_method: Callable, args: Array, hook_hash: int) -> Variant:
+	var hooks: Array = ModLoaderStore.modding_hooks.get(hook_hash, [])
+	if hooks.is_empty():
+		return vanilla_method.callv(args)
+
+	# Create a hook chain which will recursively call down until the vanilla method is reached
+	hooks.reverse()
```
pirey0: If my understanding of reverse is correct this will flip-flop the order of mod-calls every time a function is called

addons/mod_loader/internal/hooks.gd
```diff
@@ -7,29 +7,39 @@ extends Object
...
+	var chain_hook := ModLoaderHook.new(vanilla_method)
```
pirey0: I would try to get all calls in 1 ModLoaderHook instead of 1 per callback

pirey0 (changes requested):

addons/mod_loader/internal/hooks.gd
Qubus0: this was only done once when creating the chain. otherwise yes that would have been the case. not used anymore anyway tho

Alexejhero:
Are there any plans to support hooking into coroutine methods and/or having our own hooks be coroutines by awaiting execute_next?

Qubus0:
good point, i have a feeling we might currently be breaking every method that is called with await.. issue being that these methods aren't marked as async in any way so they are hard to detect with the preprocessor.
at least we can prevent breaking a whole file when we implement 
- #426

addons/mod_loader/api/hook_chain.gd
```diff
@@ -0,0 +1,48 @@
...
+## Make sure to call this method [i][color=orange]once[/color][/i] somewhere in the [param mod_callable] you pass to [method ModLoaderMod.add_hook]. [br]
+##
+## [br][b]Parameters:[/b][br]
+## - [param args] ([Array]): An array of all arguments passed into the vanilla function. [br]
+##
+## [br][b]Returns:[/b] [Variant][br][br]
+func execute_next(args := []) -> Variant:
+	_callback_index -= 1
+
+	if not _callback_index >= 0:
```
KANAjetzt:
```suggestion
	if _callback_index < 0:
```

addons/mod_loader/api/hook_chain.gd
```diff
@@ -0,0 +1,48 @@
...
+		ModLoaderLog.fatal(
+			"The hook chain index should never be negative. " +
+			"A mod hook has called execute_next twice or ModLoaderHookChain was modified in an unsupported way.",
+			LOG_NAME
+		)
+		return
+
+	var callback =  _callbacks[_callback_index]
```
KANAjetzt:
```suggestion
	var callback := _callbacks[_callback_index]
```

Is there a reason why we not want to use a specific type for the callbacks?

addons/mod_loader/api/hook_chain.gd
```diff
@@ -0,0 +1,48 @@
...
+extends RefCounted
+## Small class to keep the state of hook execution chains and move between mod hook calls.[br]
+## For examples, see [method ModLoaderMod.add_hook].
+
+
+## The reference object is usually the [Node] that the vanilla script is attached to. [br]
+## If the hooked method is [code]static[/code], it will contain the [GDScript] itself.
+var reference_object: Object
+
+var _callbacks := []
```
KANAjetzt:
```suggestion
var _callbacks: Array[Callable] = []
```

addons/mod_loader/api/hook_chain.gd
```diff
@@ -0,0 +1,48 @@
...
+var _callback_index := -1
+
+
+const LOG_NAME := "ModLoaderHookChain"
+
+
+func _init(reference_object: Object, callbacks: Array) -> void:
```
KANAjetzt, line 20:
```suggestion
func _init(reference_object: Object, callbacks: Array[Callable]) -> void:
```

KANAjetzt (commented):
Tested the provided example project in-editor and exported. Looks good 👍

addons/mod_loader/api/hook_chain.gd
Qubus0: i prefer leaving this one untyped since that is less hassle when creating a new chain (allows using `[vanilla_method] + hooks` directly without casting variables with assign again.. like done in #458)

KANAjetzt: That's worth a code comment so we can remember why it's untyped.

_Done in [7f5628a](https://github.com/GodotModding/godot-mod-loader/pull/436/commits/7f5628af2f9f34e5116ec7ade4c3e9442014629d)_

PR #440, opened 2024-11-18
feat: ✨ allow installing hooks from a file like extensions

Qubus0:
depends on #436

not quite decided what's the best way to not lose references to the files/instances that hold the hooks
1. store all script instances in a static var hashset - yet another array of references
2. force all scripts to extend Object directly, current - freefloating references that aren't cleaned up (is that necessarily bad?), no class specific methods (but are those needed? could they even be used?)
3. force all methods to be static, has no easy check - no script variables (but non static ones really needed if there is only one instance?)

test project:
[project.zip](https://github.com/user-attachments/files/17869989/project.zip)
test mod:
[mods-unpacked.zip](https://github.com/user-attachments/files/17870007/mods-unpacked.zip)


addons/mod_loader/api/mod.gd
```diff
@@ -47,6 +47,33 @@ static func install_script_extension(child_script_path: String) -> void:
...
+
+	# Every script that inherits RefCounted will be cleaned up by the engine as
+	# soon as there are no more references to it. If the reference is gone
+	# the method can't be called and everything returns null.
+	# The only exception would be static methods, but then script variables couldn't be used.
+	# If the need for other types ever arises, we can store the references too.
+	if method_holder is RefCounted:
+		ModLoaderLog.fatal("Scripts holding mod hooks should always extend Object (%s)" % hook_script_path, LOG_NAME)
+
+	var methods := script.get_script_method_list()
```
Alexejhero: It would be useful if there were some kind of way to exclude certain methods from being treated as hooks.

Qubus0: we could just explicitly filter if the vanilla script has that same method and ignore them otherwise. not sure if it's necessary though. currently everything is just added as hook, but a hook needs to be called explicitly by a vanilla method to actually do something. so if there is no method calling that hook it's almost as if it doesn't exist

Alexejhero: If hooking a method that doesn't exist doesn't cause any problems then a filter would not be necessary, you are right.

addons/mod_loader/api/mod.gd
```diff
@@ -47,6 +47,65 @@ static func install_script_extension(child_script_path: String) -> void:
...
+
+		if not OS.has_feature("editor"):
+			continue
+
+		vanilla_methods.sort_custom((
+			func(a_name: String, b_name: String, target_name: String) -> bool:
+				return a_name.similarity(target_name) > b_name.similarity(target_name)
+		).bind(hook.name))
+
+		var closest_vanilla := vanilla_methods.front()
```
Alexejhero: image
```suggestion
		var closest_vanilla: String = vanilla_methods.front()
```

KANAjetzt (approved):
Very nice! Regarding the question about not losing the reference, we already have `saved_objects` in the Store:  

https://github.com/GodotModding/godot-mod-loader/blob/baccb9c1fd832a57fcf3ec1d51ed0aaedc93a234/addons/mod_loader/mod_loader_store.gd#L83-L84

Maybe implementing something similar here would make sense? It might help with cleanup.

I tested it with the sample project I edited into the PR Description and it works 👍

PR #441, opened 2024-11-18
fix: support for Godot 4.1

Alexejhero:
### Changes made:
- Removed typed for...in variables because Godot 4.1 doesn't support them
- Downgraded some .tres to format version 3, because format version 4 isn't supported in Godot 4.1
- Used custom function instead of type_string, which does not exist in Godot 4.1

I did a proper test in Godot 4.1 and encountered no issues.

I did not test in later versions of Godot, but I don't think there's anything here that could break.

## Todos

- [x] `_ModLoaderModHookPreProcessor.fix_method_super()` erases everything after the function body, leaving only the first function in a script.
   https://github.com/GodotModding/godot-mod-loader/blob/7b36e3290e5fefa5cb26a90af85ddeabae2ebd46/addons/mod_loader/internal/mod_hook_preprocessor.gd#L313-L317
- [x] Change the ´.fix_method_super()` fix to use Regex.sub again, and just add the rest of the string back after sub replaced the super calls.


addons/mod_loader/mod_loader.gd
```diff
@@ -127,7 +127,7 @@ func _load_metadata(manifest_dict: Dictionary, zip_path := ""):
 
 func _load_mods(mod_data: Dictionary) -> void:
 	var mod_zip_paths: Array[String] = []
-	for mod: ModData in mod_data.values():
+	for mod in mod_data.values():
 		if not mod.is_loadable:
```
KANAjetzt: If we want to keep the type in the loop, we can do something fun like this:
```suggestion
	for _mod in mod_data.values():
		var mod: ModData = _mod
		if not mod.is_loadable:
```

KANAjetzt (approved):
Tested in 4.3 and works! Thank you!

Alexejhero:
So you know how they say you shouldn't make PRs from default branches? :)

Alexejhero:
Why hasn't this been merged yet? :/
I have to fix conflicts now

KANAjetzt:
I fixed the conflicts and added a fix for `_ModLoaderModHookPreProcessor.fix_method_super()` but I'm not allowed to push to your branch.

Alexejhero:
Superseded by #508

PR #456, opened 2024-11-20
fix: Errors from log.gd when using the export plugin

### LINKED ISSUE #437: [4.x] Mod Hooks: ExportPlugin should use `.process_script()` instead of `.process_script_verbose()`
image

Alexejhero:
image

ModLoaderStore exists, however none of its fields are available when being used from the tool, so `if ModLoaderStore` would return true and cause a bunch of errors

Godot version: 4.2.2

This should fix the errors mentioned in GodotModding/godot-mod-tool/pull/111

### DIFF
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -289,6 +289,9 @@ static func get_all() -> Array:
 static func get_by_mod(mod_name: String) -> Array:
 	var log_entries := []
 
+	if not _is_store_available():
+		return log_entries
+
 	if not ModLoaderStore.logged_messages.by_mod.has(mod_name):
 		error("\"%s\" not found in logged messages." % mod_name, LOG_NAME)
 		return []
@@ -349,7 +352,7 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 	if only_once and _is_logged_before(log_entry):
 		return
 
-	if ModLoaderStore:
+	if _is_store_available():
 		_store_log(log_entry)
 
 	# Check if the scene_tree is available
@@ -390,7 +393,7 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 
 
 static func _is_mod_name_ignored(mod_name: String) -> bool:
-	if not ModLoaderStore:
+	if not _is_store_available():
 		return false
 
 	var ignored_mod_names := ModLoaderStore.ml_options.ignored_mod_names_in_log as Array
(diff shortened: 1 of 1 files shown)


Qubus0:
tracked in #437 
thanks for the report though

Qubus0:
ah wait that's a pr not an issue 😅

KANAjetzt (changes requested):
Thanks for the PR!  

As suggested in the title of issue #437, we can use `.process_script()` instead of `.process_script_verbose()` in the Export Plugin to prevent these errors.

https://github.com/GodotModding/godot-mod-loader/blob/766ee3e22ea39b30e1dd5d5d2c1aca0d97df0170/addons/mod_loader/_export_plugin/export_plugin.gd#L25

Alexejhero:
> Thanks for the PR!
> 
> As suggested in the title of issue #437, we can use `.process_script()` instead of `.process_script_verbose()` in the Export Plugin to prevent these errors.
> 
> https://github.com/GodotModding/godot-mod-loader/blob/766ee3e22ea39b30e1dd5d5d2c1aca0d97df0170/addons/mod_loader/_export_plugin/export_plugin.gd#L25

While that might be true, that doesn't change the fact that the current log script improperly checks for the existence of `ModLoaderStore`. This PR fixes that issue by properly checking for its existence. This change is not mutually exclusive to the one suggested in #437

KANAjetzt:
Fair point. I need to check again why these checks are needed in the first place.
What do you think @Qubus0

Qubus0:
At this point we could even think about just adding @tool to ModLoaderLog.. there are a few instances where logs would be nice in the editor and in those cases we've already added is_silent.. if using static variables to store stuff in ModLoaderLog instead of Store works out, I think just adding tool would be the way to go. In general, not only for this issue. 

For now I'd still go without _verbose to avoid smacking the same issue with two different solutions

KANAjetzt:
Looks like the check got added during 3 to 4 conversion.
https://github.com/GodotModding/godot-mod-loader/pull/342/files

KANAjetzt:
@Alexejhero I checked again, and the additional check is only necessary if code referencing `ModLoaderStore` is executed in the editor. The export plugin should be the only component interacting with it in this way. Adding the suggested lookups for all logs seems excessive when we can simply avoid logging anything by using `.process_script()` instead of `.process_script_verbose()`.

Alexejhero:
> @Alexejhero I checked again, and the additional check is only necessary if code referencing `ModLoaderStore` is executed in the editor. The export plugin should be the only component interacting with it in this way. Adding the suggested lookups for all logs seems excessive when we can simply avoid logging anything by using `.process_script()` instead of `.process_script_verbose()`.

I only added that to 3 places, there are 3 other places that already had a check. That check that got added during 3 to 4 conversion _doesn't work_ so it should be fixed regardless of whether or not adding this to 3 extra places seems excessive.

I only added the additional check to functions that could feasibly be called in the editor, and could result in an error.

What makes you think they are excessive? Are you talking about nano-optimizations?

Alexejhero:
Superseded by #500

PR #457, opened 2024-11-20
fix: Self setup missing global classes

otDan:
- Adds all the missing classes for the self setup script

Mentioned in https://github.com/GodotModding/godot-mod-loader/issues/373


otDan:
It will need some testing, could not find a game to test this with

KANAjetzt:
Superseded by #507

PR #458, opened 2024-11-20
feat: :sparkles: support async hooks for methods using await

Qubus0:
ref https://github.com/GodotModding/godot-mod-loader/pull/436#issuecomment-2488742592

this allows using async stuff in mod hooks and applying hooks to async methods. issue being that the preprocessor needs to hook these differently and for that we need to discern if something is actually async/using await/being called with await and inserting a different hook in those cases..

```gdscript
return await _ModLoaderHooks.call_hooks_async(vanilla_1539570981_async_test_method, [param], 2690100866)
```

if we can't do it automagically, we can let the user decide it by adding `ModLoaderMod.add_async_hook()` or something similar.. but that would knock out the export plugin for the time being (though the game dev could also just add markers there like @unmoddable we have?)

we definitely want per-method hook gen
- #426

### DIFF
--- a/addons/mod_loader/api/hook_chain.gd
+++ b/addons/mod_loader/api/hook_chain.gd
@@ -1,0 +1,73 @@
+class_name ModLoaderHookChain
+extends RefCounted
+## Small class to keep the state of hook execution chains and move between mod hook calls.[br]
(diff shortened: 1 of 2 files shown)


KANAjetzt:
closed because of history issues move to #470

PR #474, opened 2024-11-27
fix: :ambulance: remove whitespace after `_ModLoaderHooks.call_hooks`

KANAjetzt:
Breaks my test code every time I save in the editor.


Qubus0 (approved):
no need to add trailing whitespace anyway 👍

PR #476, opened 2024-11-28
test: :test_tube: added tests for `.process_script()`

KANAjetzt:
This adds 3 test scripts (`test_script_A / B / C.gd`) with their expected results (`test_script_A / B/ C_processed.gd`) to validate the functionality of `_ModLoaderModHookPreProcessor.process_script()`.

> [!NOTE]
>  The `class_name` in the `_processed.gd` scripts must be commented out to avoid the "hiding global class" error.
> Test run is broken until #495 is merged or `func please_stop` is removed from test script C.

## TODO

- [x] #494 

---

<details><summary>Dev Notes</summary>
<p>

- add this to the test script 
```gdscript
format_date(day: int, \
month: int, year: int, special_thing := "\n\t\"\
")
```
- I think this covers it quiet well already:
```gdscript
func why_would_you(put: int, \
	backslashes := "\\", in_here := "?!\n"
	):
	pass
```

</p>
</details>

### DIFF
--- a/test/Unit/test_mod_hook_preprocessor.gd
+++ b/test/Unit/test_mod_hook_preprocessor.gd
@@ -95,9 +95,9 @@ func test_process_script() -> void:
 	var hook_pre_processor := _ModLoaderModHookPreProcessor.new()
 	hook_pre_processor.process_begin()
 
-	var result_a := hook_pre_processor.process_script("res://test_mod_hook_preprocessor/test_script_A.gd")
+	var result_a := hook_pre_processor.process_script("res://test_mod_hook_preprocessor/test_script_A.gd", false)
 	var result_a_expected: String = load("res://test_mod_hook_preprocessor/test_script_A_processed.gd").source_code.trim_prefix("#")
-	var result_b := hook_pre_processor.process_script("res://test_mod_hook_preprocessor/test_script_B.gd")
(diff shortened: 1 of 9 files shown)


test/Unit/test_mod_hook_preprocessor.gd
```diff
@@ -89,3 +89,19 @@ func test_match_func_with_whitespace(params: Array = use_parameters(test_match_f
...
 		"expected %s, got %s" % [expected_string, result.get_string()]
 	)
+
+
+func test_process_script() -> void:
+	var hook_pre_processor := _ModLoaderModHookPreProcessor.new()
+	hook_pre_processor.process_begin()
+
+	var result_a := hook_pre_processor.process_script("res://test_mod_hook_preprocessor/test_script_A.gd")
+	var result_a_expected: String = load("res://test_mod_hook_preprocessor/test_script_A_processed.gd").source_code.trim_prefix("#")
```
Qubus0, line 100: `.trim_prefix("#")` smart strategy to avoid the double class name, might need a short comment so we don't forget tho

Qubus0 (approved):
nice. this must be the funniest code in the whole repo.

i think the long term goal is to test all the cases separately, each with their own script. that way we'll know exactly what failed at a glance. but finding which one causes issues is mostly possible already and we also get tests for things like duplicate names and just multiple variables and functions in the same file in general. 

one small addition would be subclasses to the uncovered cases

PR #482, opened 2024-12-05
fix: :bug: Take resource remapping into account

### LINKED ISSUE #481: use `ResourceLoader.exists()` to prevent ERROR: A critical file is missing: res://addons/mod_loader/options/options.tres
https://github.com/GodotModding/godot-mod-loader/blob/1bac6f664029dd2f494683635717cdb67aebbd87/addons/mod_loader/mod_loader_store.gd#L152-L154

https://docs.godotengine.org/en/stable/classes/class_resourceloader.html#class-resourceloader-method-exists

KANAjetzt:
added `ResourceLoader.exists()` to `_ModLoaderFile.file_exists()`

closes #481

### DIFF
--- a/addons/mod_loader/internal/file.gd
+++ b/addons/mod_loader/internal/file.gd
@@ -196,7 +196,13 @@ static func file_exists(path: String, zip_path: String = "") -> bool:
 	if not zip_path.is_empty():
 		return file_exists_in_zip(path, zip_path)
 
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
yep that also works. one method to rule them all instead of always thinking about if it's a resource or not

PR #491, opened 2024-12-16
feat: mod loader version check

Qubus0:
leftover todo from #416

helps prevent crashes when the mod loader was updated (to support hooks for example) but the game did not receive a major update.


Qubus0:
hm this might have to be rethought since the mod loader should be downward compatible while we assume that the game is not 🤔

KANAjetzt:
Maybe always display a warning if a mod loader version mismatch is detected.

Alternatively, warn for a major version difference of -1 and disable the mod for a major version difference of -2, since we say that deprecated methods are removed in the next major version.

PR #493, opened 2024-12-21
fix: :bug: check if a method is already hooked

KANAjetzt:
Added a check to `_ModLoaderModHookPreProcessor.process_script()` to determine if a method is already hooked by checking if the `method.name` starts with the prefix added by the preprocessor.

> [!NOTE]
> It will create an empty zip if no new dynamic hooks are created.


addons/mod_loader/internal/mod_hook_packer.gd
```diff
@@ -50,6 +50,11 @@ static func start() -> void:
 
 		var processed_source_code := hook_pre_processor.process_script_verbose(path)
 
+		# Skip writing to the zip if no new hooks were created for this script
+		if not hook_pre_processor.script_paths_hooked.has(path):
+			ModLoaderLog.debug("No new hooks where created in \"%s\" skipping writing to hook pack." % path, LOG_NAME)
```
Qubus0: *were created

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -65,10 +66,24 @@ func process_script(path: String, enable_hook_check := false) -> String:
 		is_func_moddable.bind(source_code, getters_setters)
 	)
 
+	var methods_hooked := {}
+
 	for method in moddable_methods:
 		if method.name in method_store:
 			continue
```
Qubus0, line 73: does this new way replace method_store?

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -65,10 +66,24 @@ func process_script(path: String, enable_hook_check := false) -> String:
...
 
+		var prefix := "%s%s_" % [METHOD_PREFIX, class_prefix]
+
+		# Verifies if the method has already been processed and hooked before
+		# to avoid duplicate hooks on the same function.
+		if method.name.begins_with(prefix):
```
Qubus0, line 80: this is essentially checking the script method list for methods that begin with something like vanilla_123123 - which are the all automatically generated by us, right? took me a bit to get, a bit more detail in the comment would be nice

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -65,10 +66,24 @@ func process_script(path: String, enable_hook_check := false) -> String:
...
+			var method_name_vanilla: String = method.name.trim_prefix(prefix)
+			methods_hooked[method_name_vanilla] = true
+			continue
+
+		if methods_hooked.has(method.name):
+			continue
```
Qubus0, line 88: this takes effect once a vanilla method was added to the list, right? the order which methods are iterated through should be alphabetical right? which means the vanilla_ methods would be pretty late in the chain, so the actual vanilla method would be checked before it is added to the list. is that right?

so we can either sort the vanilla_ methods to come first or loop twice. i guess it's looping twice in either case

Qubus0: ah looks like the order is the same as the declaration order, not alphabetical. since we move the vanilla names to the end they will always be last and included 👍 

```
vanilla_1539570981__ready
vanilla_1539570981_await_method
vanilla_1539570981_greeting
vanilla_1539570981_async_test_method
vanilla_1539570981_wait_await
_ready
await_method
greeting
async_test_method
wait_await
```

addons/mod_loader/internal/mod_hook_preprocessor.gd
KANAjetzt: No - `method_store` is because get_script_method_list() also returns methods of a potential parent class.

https://github.com/GodotModding/godot-mod-loader/blob/aab3aaa089f75404f07dd790aa193bb430a5582a/addons/mod_loader/internal/mod_hook_preprocessor.gd#L120-L125

addons/mod_loader/internal/mod_hook_preprocessor.gd
KANAjetzt: Exactly. I hope the new comment is clearer. Feel free to make edits.
_[5c3e32e](https://github.com/GodotModding/godot-mod-loader/pull/493/commits/5c3e32ed27606e490a7a0a90b0472607ceaf0672)_

addons/mod_loader/internal/mod_hook_packer.gd
KANAjetzt: ty - _[999b71c](https://github.com/GodotModding/godot-mod-loader/pull/493/commits/999b71c75263cc6b19382479f220a15833a8b638)_

Qubus0 (approved):

PR #498, opened 2024-12-24
fix: :bug: resolved parser error because of missing return value

### LINKED ISSUE #437: [4.x] Mod Hooks: ExportPlugin should use `.process_script()` instead of `.process_script_verbose()`

### LINKED ISSUE #487: skip / remove ` ModLoaderStore.get("any_mod_hooked")` check in export plugin
>               I think the` ModLoaderStore.get("any_mod_hooked")` kills all the performance-benefits of checking ModLoaderStore.any_mod_hooked. 
> This was originally a way to reduce the modding overhead when running a modless version of a game. 
> 2 topics:
>  - can we somehow skip the ` ModLoaderStore.get("any_mod_hooked")`  call?
>  - can we skip the check altogether when dynamically generating the scripts, since there we only create the methods when hooks actually exist.
>  
>  (Same below on l.271)
> 
> _Originally posted by @pirey0 in https://github.com/GodotModding/godot-mod-loader/pull/408#discussion_r1832930546_

> [!NOTE]
>   #429 Removed this check from dynamic hook creation.

In addition to the mentioned performance issue, we currently have the following Parse error:
`SCRIPT ERROR: Parse Error: Not all code paths return a value.`

Example:
```gdscript
func check_sight(target_position: Vector3) -> bool:
	if ModLoaderStore.get("any_mod_hooked") and ModLoaderStore.any_mod_hooked:
		return _ModLoaderHooks.call_hooks(vanilla_608418131_check_sight, [target_position], 1845897103)
```

KANAjetzt:
Added an else statement that directly calls or returns the vanilla function if `any_mod_hooked` is `false`.

Result:
```gdscript
func _ready():
	if ModLoaderStore.any_mod_hooked:
		_ModLoaderHooks.call_hooks(vanilla_3694473935__ready, [], 1946934147)
	else:
		vanilla_3694473935__ready()


func play_global_sfx(audio: AudioStream, db: float=0.0, delay: =0.0, pitch_scale: float=randf_range(0.85, 1.2)):
	if ModLoaderStore.any_mod_hooked:
		await _ModLoaderHooks.call_hooks_async(vanilla_2851983457_play_global_sfx, [audio, db, delay, pitch_scale], 2151934167)
	else:
		await vanilla_2851983457_play_global_sfx(audio, db, delay, pitch_scale)


func check_sight(target_position: Vector3) -> bool:
	if ModLoaderStore.any_mod_hooked:
		return _ModLoaderHooks.call_hooks(vanilla_608418131_check_sight, [target_position], 1845897103)
	else:
		return vanilla_608418131_check_sight(target_position)
```

closes #487
closes #437

### DIFF
--- a/addons/mod_loader/_export_plugin/export_plugin.gd
+++ b/addons/mod_loader/_export_plugin/export_plugin.gd
@@ -22,6 +22,6 @@ func _export_file(path: String, type: String, features: PackedStringArray) -> vo
 	skip()
 	add_file(
 		path,
-		hook_pre_processor.process_script_verbose(path, true).to_utf8_buffer(),
+		hook_pre_processor.process_script(path, true).to_utf8_buffer(),
 		false
 	)
--- a/addons/mod_loader/internal/mod_hook_preprocessor.gd
+++ b/addons/mod_loader/internal/mod_hook_preprocessor.gd
@@ -354,11 +354,15 @@ static func build_mod_hook_string(
 	var async_string := "_async" if is_async else ""
 	var return_var := "var %s = " % "return_var" if not method_type.is_empty() or return_prop_usage == 131072 else ""
 	var method_return := "return " if not method_type.is_empty() or return_prop_usage == 131072 else ""
-	var hook_check := 'if ModLoaderStore.get("any_mod_hooked") and ModLoaderStore.any_mod_hooked:\n\t\t' if enable_hook_check else ""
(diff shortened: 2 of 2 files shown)


Qubus0 (approved):
doesn't need to be an else since we already return in the if. both are fine either way tho
```gdscript
if hooked:
    return modloaderhooks..
return vanilla
```

Issue #452, opened 2024-11-18
phase out the ModLoaderStore completely by using static variables


Qubus0:
would be a lot cleaner, more stable since we have one less autoload to worry about and set up and be more specific about where these variables are stored.

KANAjetzt:
Just to add it here, one downside of this would be losing easy access to all data via the Remote Tab in the Scene Tree view. I don't think there is a way to get a debug view without setting a breakpoint?

Not a deal breaker, just an inconvenience.

KANAjetzt:
> [Qubus0](https://github.com/Qubus0) [3 days ago](https://github.com/GodotModding/godot-mod-loader/pull/500#discussion_r1897885905)
> > Also static vars make it harder to debug because you can't simply view them in the Remote view.
> 
> i haven't made much use of that but good point. but since we can actually inspect them like this, if it ever gets too annoying, we can just keep the store around purely for inspecting these values. probably renamed to something else like `StaticInspector` and not mandatory as autoload. or as a child of the mod loader when in the editor, something like that
> 
> ```gdscript
> var logged_messages: Dictionary:
> 	get:
> 		return ModLoaderLog.logged_messages
> ```
https://github.com/GodotModding/godot-mod-loader/pull/500#discussion_r1897885905

Qubus0:
static variables were added in 4.1, so we might be locking ourselves out from 4.0 games. so far there hasn't been a major game on our radar that used just 4.0 though

ref https://github.com/godotengine/godot/pull/76264

PR #499, opened 2024-12-25
Feat: logging without store

Qubus0:
Removes all variables from ModLoaderStore to ModLoaderLog - as static variables
also added `@tool` since there are a few cases popping up that make use of the logger in editor (like the mod tool)

that should also let us remove `is_silent` from a few functions like `is_name_or_namespace_valid` if we want to.

supersedes and closes #456

works towards 
- #452


Qubus0:
replaced by #500

PR #500, opened 2024-12-25
Feat: Logging without store

Qubus0:
Moves all logging variables from `ModLoaderStore` to `ModLoaderLog` - as static variables
also added `@tool` since there are a few cases popping up that make use of the logger in editor (like the mod tool)

that should also let us remove `is_silent` from a few functions like `is_name_or_namespace_valid` if we want to.

supersedes #456

works towards 
- #452

### DIFF
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -8,18 +8,19 @@ extends Object
 # Path to the latest log file.
 const MOD_LOG_PATH := "user://logs/modloader.log"
 
-const LOG_NAME := "ModLoader:Log"
+const _LOG_NAME := "ModLoader:Log"
 
+## Denotes the severity of a log entry
 enum VERBOSITY_LEVEL {
-	ERROR,
-	WARNING,
-	INFO,
-	DEBUG,
+	ERROR, 		## For errors and fatal errors
+	WARNING, 	## For warnings
+	INFO, 		## For everything informational and successes
+	DEBUG, 		## For debugging, can get quite verbose
 }
 
-# Keeps track of logged messages, to avoid flooding the log with duplicate notices
-# Can also be used by mods, eg. to create an in-game developer console that
-# shows messages
+## Keeps track of logged messages, to avoid flooding the log with duplicate notices
+## Can also be used by mods, eg. to create an in-game developer console that
+## shows messages
 static var logged_messages := {
 	"all": {},
 	"by_mod": {},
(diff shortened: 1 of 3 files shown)


addons/mod_loader/api/log.gd
```diff
@@ -16,6 +17,28 @@ enum VERBOSITY_LEVEL {
 	DEBUG,
 }
 
+# Keeps track of logged messages, to avoid flooding the log with duplicate notices
+# Can also be used by mods, eg. to create an in-game developer console that
+# shows messages
```
KANAjetzt: While we are here, we can turn these into doc comments.

addons/mod_loader/mod_loader_store.gd
```diff
@@ -92,21 +92,13 @@ var saved_mod_mains := {}
...
-		"warning": {},
-		"info": {},
-		"success": {},
-		"debug": {},
-	}
-}
+var logged_messages: Dictionary:
+	set(val):
+		ModLoaderDeprecated.deprecated_message("ModLoaderStore.logged_messages was moved to ModLoaderLog.logged_messages", "7.0.1")
+		ModLoaderLog.logged_messages = val
```
KANAjetzt: Is there "only log once" for `deprecated_message()`?
If you currently click on in the Editor in the Remote Scene View on ModLoader you get spammed with:
`WARNING ModLoader:Deprecated: DEPRECATED: ModLoaderStore.logged_messages was moved to ModLoaderLog.logged_messages (since version 7.0.1)`

KANAjetzt: Also static vars make it harder to debug because you can't simply view them in the Remote view.

addons/mod_loader/api/log.gd
```diff
@@ -375,65 +397,56 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
...
 		"debug":
-			if _get_verbosity() >= VERBOSITY_LEVEL.DEBUG:
+			if verbosity >= VERBOSITY_LEVEL.DEBUG:
 				print(log_entry.get_prefix() + message)
 				_write_to_log_file(log_entry.get_entry())
 
 
 static func _is_mod_name_ignored(mod_name: String) -> bool:
-	if not ModLoaderStore:
+	if ignored_mods.size() == 0:
```
KANAjetzt:
```suggestion
	if ignored_mods.is_empty():
```

KANAjetzt (approved):
very nice - will be great for the dev tool.

addons/mod_loader/mod_loader_store.gd
Qubus0: might actually make sense to make all deprecated errors `only_once`

Qubus0: > Also static vars make it harder to debug because you can't simply view them in the Remote view.

i haven't made much use of that but good point. but since we can actually inspect them like this, if it ever gets too annoying, we can just keep the store around purely for inspecting these values. probably renamed to something else like `StaticInspector` and not mandatory as autoload. or as a child of the mod loader when in the editor, something like that
```gdscript
var logged_messages: Dictionary:
	get:
		return ModLoaderLog.logged_messages
```

PR #508, opened 2025-01-04
refactor: ♻️ 4.1 support

KANAjetzt:
supersedes #441

Original PR Description:

> ### Changes made:
> - Removed typed for...in variables because Godot 4.1 doesn't support them
> - Downgraded some .tres to format version 3, because format version 4 isn't supported in Godot 4.1
> - Used custom function instead of type_string, which does not exist in Godot 4.1
> 
> I did a proper test in Godot 4.1 and encountered no issues.
> 
> I did not test in later versions of Godot, but I don't think there's anything here that could break.
> 
> ## Todos
> 
> - [x] `_ModLoaderModHookPreProcessor.fix_method_super()` erases everything after the function body, leaving only the first function in a script.
>    https://github.com/GodotModding/godot-mod-loader/blob/7b36e3290e5fefa5cb26a90af85ddeabae2ebd46/addons/mod_loader/internal/mod_hook_preprocessor.gd#L313-L317
> - [x] Change the ´.fix_method_super()` fix to use Regex.sub again, and just add the rest of the string back after sub replaced the super calls.

## Todos

- [x] Godot 4.2 Test (Tested with Dome Keeper)
- [x] Godot 4.3 Test (Tested with [flashit](https://github.com/KANAjetzt/MFGJ-S-24-Flashing))


addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -13,15 +14,18 @@ const HASH_COLLISION_ERROR := \
...
 ## \\b%s\\b 			->	the function name
 ## (?:.*\\n*)*?\\s*\\( 	->	Match any character between zero and unlimited times, but be lazy
 ## 							and only do this until a '(' is found.
 const REGEX_MATCH_FUNC_WITH_WHITESPACE := "\\bfunc\\b\\s+\\b%s\\b(?:.*\\n*)*?\\s*\\("
 
+static var engine_version_hex: int = Engine.get_version_info().hex
+
 ## finds function names used as setters and getters (excluding inline definitions)
-## group 2 and 4 contain the xetter names
+## group 2 and 4 contain the setter names
```
Qubus0: both setter and getter names (x-etter would probably have been a better spelling. or *etter)

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -495,7 +520,7 @@ static func get_return_type_string(return_data: Dictionary) -> String:
 	if return_data.has("class_name") and not str(return_data.class_name).is_empty():
 		type_base = str(return_data.class_name)
 	else:
-		type_base = type_string(return_data.type)
+		type_base = get_type_name(return_data.type)
```
Qubus0, line 523: let's also use an `engine_version_hex` check here and only use our custom one for 4.1 and the engine version for everything else - that way we'll be safe in future versions if they every add another type for any reason

Qubus0 (approved):
looks good, ty.
the resources might get annoying since they will always want to be saved in the newer format version. 
we might also want to consider an engine version test that just runs `--check-only` on each of our scripts for every version we want to support in order to avoid accidentally using things like typed loop variables.

addons/mod_loader/internal/mod_hook_preprocessor.gd
KANAjetzt: Maybe like this?

```suggestion
## group 2 and 4 contain the setter/getter names
```

Qubus0: ye

addons/mod_loader/internal/mod_hook_preprocessor.gd
KANAjetzt: Doesn't work, we get a parse error because `type_string()` doesn't exist in 4.1 .

Qubus0: right, we'd need to move it to a new file, something like a version adapter

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -535,3 +560,90 @@ static func get_hook_check_else_string(
...
+		TYPE_PACKED_VECTOR2_ARRAY:
+			return "PackedVector2Array"
+		TYPE_PACKED_VECTOR3_ARRAY:
+			return "PackedVector3Array"
+		TYPE_PACKED_COLOR_ARRAY:
+			return "PackedColorArray"
+		38: # TYPE_PACKED_VECTOR4_ARRAY
+			return "PackedVector4Array"
+	push_error("Argument `type` is invalid. Use `TYPE_*` constants.")
+	return "<invalid type>"
```
Qubus0:
```suggestion
	return "<unknown type %s>" % type
```
okay i think we can merge without the engine method and see if it's even an issue. but just make debugging easier i case it's a type we don't know, let's include the id in the message

PR #515, opened 2025-01-14
fix: :bug: return detection

KANAjetzt:
<details><summary>Old Description</summary>
<p>

The current return detection relies on data from `method_list`, but if a method does not have a statically typed return value, the `return_prop_usage == 131072` check does not appear to catch all return cases. This fix introduces `is_method_returning`, which checks the `func_body_string` for an uncommented `return`.

TODO:

- [x] Make sure the return string a single word
   - [x] without ignoring returns on the first or last line

</p>
</details> 

We can always return ?!
 - No 
image
- It seems we can always `return` except in `_init()`.

### DIFF
--- a/addons/mod_loader/internal/mod_hook_preprocessor.gd
+++ b/addons/mod_loader/internal/mod_hook_preprocessor.gd
@@ -88,6 +88,7 @@ func process_script(path: String, enable_hook_check := false) -> String:
 			continue
 
 		var type_string := get_return_type_string(method.return)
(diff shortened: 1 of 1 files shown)


Qubus0 (approved):
i'd still generalize `is_constructor` to `can_return` and also have a check for `-> void` being present. it's currently completely fine to return a method even if that method is statically typed (see image). but i would assume this will be fixed at some point, so we should not neglect it. 
image
topic for a follow up pr

PR #518, opened 2025-01-24
fix: :bug: check for `-> void`

### LINKED ISSUE #517: Generalize `is_constructor` to `can_return` and also have a check for `-> void`
i'd still generalize `is_constructor` to `can_return` and also have a check for `-> void` being present. it's currently completely fine to return a method even if that method is statically typed (see image). but i would assume this will be fixed at some point, so we should not neglect it. 
image
topic for a follow up pr

_Originally posted by @Qubus0 in https://github.com/GodotModding/godot-mod-loader/pull/515#pullrequestreview-2556815213_

KANAjetzt:
prevents parse errors if hook check is enabled and is cleaner in general.

closes #517

### DIFF
--- a/addons/mod_loader/internal/mod_hook_preprocessor.gd
+++ b/addons/mod_loader/internal/mod_hook_preprocessor.gd
@@ -456,12 +456,13 @@ static func is_comment(text: String, start_index: int) -> bool:
 	return true
 
 
+# Get the left side substring of a line from a given start index
 static func get_line_left(text: String, start: int) -> String:
-	var left := text.substr(0, start)
-	var line_start_index := left.rfind("\n", start) + 1
-	return left.right(left.length() - line_start_index)
+	var line_start_index := text.rfind("\n", start) + 1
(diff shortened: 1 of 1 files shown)


addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -115,7 +116,11 @@ func process_script(path: String, enable_hook_check := false) -> String:
 		if not func_def: # If no valid function definition is found after processing.
 			continue # Skip to the next iteration.
 
-		var func_body_start_index := get_func_body_start_index(func_def.get_end(), source_code)
+		# Shift the func_def_end index back by one to start on the opening parentheses.
+		# Because the match_func_with_whitespace().get_end() is the index after the opening parentheses.
+		var closing_paran_index := get_closing_paren_index(func_def.get_end() - 1, source_code)
```
Qubus0: typo paran

Qubus0 (approved):
nice, thank you ✨

PR #523, opened 2025-01-25
fix: :bug: set `current_config` in `ModData._load_config()`

### LINKED ISSUE #466: Mod Configs:  `current_config` was `null` because it isn't set in `ModData._load_config`
Can be salvaged from 
- #394 

> - Fixed an issue where `current_config` was `null` because it wasn't set in `ModData._load_config`.

https://github.com/KANAjetzt/godot-mod-loader/blob/cea915e276a59c7b8c89f4c103e7d3147fdbf879/addons/mod_loader/resources/mod_data.gd#L125-L127

KANAjetzt:
Fixed an issue where `current_config` was null because it wasn't set in `ModData._load_config()`.

closes #466

### DIFF
--- a/addons/mod_loader/resources/mod_data.gd
+++ b/addons/mod_loader/resources/mod_data.gd
@@ -104,6 +104,10 @@ func _load_config(config_file_path: String) -> void:
 	# Add the config to the configs dictionary
 	configs[mod_config.name] = mod_config
 
+	# Set it as the current_config if there is none
+	if not current_config:
+		current_config = mod_config
+
 
 # Update the mod_list of the current user profile
 func _set_current_config(new_current_config: ModConfig) -> void:
--- a/addons/mod_loader/resources/mod_data.gd
+++ b/addons/mod_loader/resources/mod_data.gd
@@ -88,7 +88,8 @@ func load_configs() -> void:
 		_load_config(config_file_path)
 
 	# Set the current_config based on the user profile
-	current_config = ModLoaderConfig.get_current_config(dir_name)
+	if ModLoaderUserProfile.is_initialized():
+		current_config = ModLoaderConfig.get_current_config(dir_name)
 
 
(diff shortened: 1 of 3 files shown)


addons/mod_loader/resources/mod_data.gd
```diff
@@ -104,6 +104,10 @@ func _load_config(config_file_path: String) -> void:
 	# Add the config to the configs dictionary
 	configs[mod_config.name] = mod_config
 
+	# Set it as the current_config if there is none
+	if not current_config:
+		current_config = mod_config
```
Qubus0: there's no priority for which config will be loaded first right? so this basically assigns a random one as current, which may be unexpected. still better than none at all.

though i feel like this is a better fit somewhere outside this load function - since i would expect it to not have any side effects. probably just after `_load_config` is called

Qubus0 (approved):
works for now, just some extra thoughts

addons/mod_loader/resources/mod_data.gd
```diff
@@ -88,7 +88,8 @@ func load_configs() -> void:
 		_load_config(config_file_path)
 
 	# Set the current_config based on the user profile
-	current_config = ModLoaderConfig.get_current_config(dir_name)
+	if ModLoaderUserProfile.is_initialized():
+		current_config = ModLoaderConfig.get_current_config(dir_name)
```
Qubus0, line 92: would we set the default config if it's not initialized yet? or does that happen somewhere else?

Qubus0 (approved):
that's nicer

addons/mod_loader/resources/mod_data.gd
KANAjetzt: yes ty 😄 - [4d19069](https://github.com/GodotModding/godot-mod-loader/pull/523/commits/4d1906921b02591dcf7864775ed4a2ee7d4f7e94)

PR #529, opened 2025-01-26
docs: Update README.md

KANAjetzt:
We can remove the `class_name` restriction.


Qubus0 (approved):

Exerionius:
1. If mentions of `class_name` limitations were removed, does it mean that the issue was resolved?
2. If the issue isn't resolved but there's a workaround, is it documented somewhere? Besides "just don't use class names".
3. How it relates to this? https://github.com/godotengine/godot/issues/83542

Qubus0:
Hey @Exerionius, the `class_name` issue persists, but we've worked around it with [script hooks](https://github.com/GodotModding/godot-mod-loader/pull/408), which is essentially runtime modification and patching of scripts. Doc update still work in progress, but you can see some drafts here https://qubus0.github.io/gmlwiki/guides/modding/script_hooks/

Issue #468, opened 2024-11-23
Demote "No profile" error from error to info in `ModLoaderUserProfile`


KANAjetzt:
From #394
> - Demoted "No profile" error from error to info because on the first run of the game, the file will not be there.

https://github.com/GodotModding/godot-mod-loader/blob/6a3d76e5749e03517f6cadea1c522108760270ff/addons/mod_loader/api/profile.gd#L415

KANAjetzt:
closed by https://github.com/GodotModding/godot-mod-loader/pull/530

PR #530, opened 2025-01-28
fix: :bug: moved default profile creation to `_init()`

### LINKED ISSUE #465: Move default profile creation in mod_loader.gd from `_ready()` to `_load_mods()`
Can be salvaged from 
- #394 

> Moved default profile creation in mod_loader.gd from `_ready()` to `_load_mods()`.
> This is required so mod configs can be loaded on the first run of the game.

https://github.com/KANAjetzt/godot-mod-loader/blob/cea915e276a59c7b8c89f4c103e7d3147fdbf879/addons/mod_loader/mod_loader.gd#L111-L121

The default user profile should be created before mod configs are loaded. Without a user profile, the configs cannot be saved.

KANAjetzt:
Added `ModLoaderUserProfile.is_initialized()` and a fallback to the default config in `ModData.load_configs()`. Moved default user profile creation and user profile `mod_list` update to the mod loader `_init()`. With these changes, all config and user profile data is available on the first run.

closes #465 #468

### DIFF
--- a/addons/mod_loader/api/profile.gd
+++ b/addons/mod_loader/api/profile.gd
@@ -199,6 +199,13 @@ static func get_all_as_array() -> Array:
 	return user_profiles
 
 
+# Returns true if the Mod User Profiles are initialized.
+# On the first execution of the game, user profiles might not yet be created.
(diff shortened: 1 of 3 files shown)


Qubus0 (approved):
nice

PR #534, opened 2025-01-30
feat: per method hooks

### LINKED ISSUE #426: dynamic hooking: selectively only hook targeted methods instead of whole file

Qubus0:
closes #426 

Hooks are now applied by method instead of by script. 

This should make hooks safer in general - even if one method may not be processed correctly, it won't break the whole script or prevent it from being modded.

caching of the methods is slightly different now: if there is any difference in which scripts or methods are hooked, everything is processed. This is a bit less efficient, but also more predictable while still reducing the processing required if the game is started with the exact same mods/hooks, and reducing required restarts because of it


addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -51,19 +51,17 @@ func process_begin() -> void:
 	hashmap.clear()
 
 
-func process_script_verbose(path: String, enable_hook_check := false) -> String:
+func process_script_verbose(path: String, enable_hook_check := false, method_mask: Array[String] = []) -> String:
```
KANAjetzt, line 54: Is `method_mask` = methods that need to be hooked?

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -72,32 +70,35 @@ func process_script(path: String, enable_hook_check := false) -> String:
...
 			methods_hooked[method_name_vanilla] = true
 			continue
-
 		# This ensures we avoid creating a hook for the 'imposter' method, which
 		# is generated by `build_mod_hook_string()` and has the vanilla method name.
 		if methods_hooked.has(method.name):
 			continue
 
+		# If a mask is provided, only methods with their name in the mask will be converted.
+		# Can't be pre-filtered since it removes prefixed methods required by the previous check.
```
KANAjetzt: `# Can't be pre-filtered since it removes prefixed methods required by the previous check.`
I don't understand that comment 😄 

do you mean pre-filtered before the script is processed?

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -400,7 +401,7 @@ static func build_mod_hook_string(
 
 	return """
 {STATIC}func {METHOD_NAME}({METHOD_PARAMS}){RETURN_TYPE_STRING}:
-	{HOOK_CHECK}{RETURN}{AWAIT}_ModLoaderHooks.call_hooks{ASYNC}({METHOD_PREFIX}_{METHOD_NAME}, [{METHOD_ARGS}], {HOOK_ID}){HOOK_CHECK_ELSE}
+	{HOOK_CHECK}{RETURN}{AWAIT}_ModLoaderHooks.call_hooks{ASYNC}({METHOD_PREFIX}{METHOD_NAME}, [{METHOD_ARGS}], {HOOK_ID}){HOOK_CHECK_ELSE}
```
KANAjetzt, line 406: removed the `_` here in 9ad58c0878530314b0ca6ed6369ba61869093598
It's now part of `full_prefix` / `METHOD_PREFIX`

addons/mod_loader/internal/mod_hook_preprocessor.gd
```diff
@@ -551,7 +552,7 @@ static func get_hook_check_else_string(
 	method_name: String,
 	method_arg_string_names_only: String
 ) -> String:
-	return "\n\telse:\n\t\t{RETURN}{AWAIT}{METHOD_PREFIX}_{METHOD_NAME}({METHOD_ARGS})".format(
+	return "\n\telse:\n\t\t{RETURN}{AWAIT}{METHOD_PREFIX}{METHOD_NAME}({METHOD_ARGS})".format(
```
KANAjetzt, line 557: removed the `_` here in 9ad58c0878530314b0ca6ed6369ba61869093598
It's now part of `full_prefix` / `METHOD_PREFIX`

KANAjetzt (approved):
nice - ty 👍

addons/mod_loader/internal/mod_hook_preprocessor.gd
Qubus0: before the loop essentially, like it was before the fix, similar to the is_func_moddable filter. updated the comment slightly

addons/mod_loader/internal/mod_hook_preprocessor.gd
Qubus0: yep. not sure if another variable name like method_filter makes a lot more sense, so i just added some doc comments, which were in order anyway

PR #536, opened 2025-01-31
feat: ✨ options for game version validation

KANAjetzt:
### Game Version Validation

```gdscript
enum VERSION_VALIDATION {
   DEFAULT,
   DISABLED,
   CUSTOM,
}
```

Choose between:

- **DEFAULT:** Uses standard semantic versioning (semver) validation.
- **DISABLED:** Completely disables version validation.
- **CUSTOM:** Allows for custom validation logic.

### Customize Script Path
```gdscript
@export_file var customize_script_path: String
```

To apply settings that cannot be configured through the editor UI.
  
Currently, it is required for custom game version validation.
The custom validation uses `custom_game_version_validation_callable` a `Callable` stored in the `ml_options`.

I added an example script here:
https://github.com/KANAjetzt/godot-mod-loader/blob/feat_setting-to-disable-version-check-or-use-custom-one/addons/mod_loader/options/example_customize_script.gd

I'm open to suggestions or alternative solutions that require less setup work.


addons/mod_loader/resources/options_profile.gd
```diff
@@ -1,18 +1,74 @@
...
+	## and the mod's [member ModManifest.compatible_game_version].
+	DISABLED,
+
+	## Enables custom game version validation.
+	## Use [member customize_script_path] to specify a script that customizes the Mod Loader options.
+	## In this script, you must set [member custom_game_version_validation_callable]
+	## to a custom validation [Callable].
+	##
+	## Example:
+	## [codeblock]
```
Qubus0: i think we can either use this doc comment or the example script and point to it, but we don't need both at once. or perhaps move it to tests

Qubus0 (approved):
it's a little weird, but it's certainly way better than not having it at all 😄

addons/mod_loader/resources/options_profile.gd
KANAjetzt: the example script to tests? 
We can also move it to the docs page and just use `@tutorial`.

Qubus0: Yeah turn it into a test and use that test as example. But true, docs are probably best.

addons/mod_loader/resources/options_profile.gd
```diff
@@ -2,6 +2,8 @@ class_name ModLoaderOptionsProfile
 extends Resource
 ##
 ## Class to define and store Mod Loader Options.
+##
+## @tutorial(Example Customization Script): https://example.com
```
Qubus0: i think we can preemptively link to `https://wiki.godotmodding.com/guides/integration/mod_loader_options/#game-version-validation`
which would just be linking to a heading like this https://qubus0.github.io/gmlwiki/guides/integration/mod_loader_options/#available-options
the updated docs will be up with the release anyway

addons/mod_loader/resources/options_profile.gd
```diff
@@ -17,36 +19,12 @@ enum VERSION_VALIDATION {
...
-	##            )
-	##            return false
-	##
-	##    return true
-	## [/codeblock]
-	##
+	## [br]
+	## ===[br]
+	## [b]Note:[color=note "Easier Mod Loader Updates"][/color][/b][br]
 	## Using a customization script allows you to keep your custom code outside the addon directory,
```
Qubus0:
```suggestion
	## Using a custom script allows you to keep your code outside the addons directory,
```

Qubus0 (approved):
nice, thanks ✨

addons/mod_loader/resources/options_profile.gd
```diff
@@ -1,18 +1,52 @@
 class_name ModLoaderOptionsProfile
 extends Resource
+##
+## Class to define and store Mod Loader Options.
+##
+## @tutorial(Example Customization Script): https://example.com
```
KANAjetzt:
```suggestion
## @tutorial(Example Customization Script): https://wiki.godotmodding.com/guides/integration/mod_loader_options/#game-version-validation
```

PR #538, opened 2025-02-01
feat: new editor-only log type for dev hints

Qubus0:
small improvement to highlight specific log messages in the editor meant only for developers. these are not logged in the exported project to keep the file clean.

i've also demoted the one mod hook warning to a hint since the only reason for it being a warning was to draw attention to it, which the hint does better. We should never leave a stray warning in the debugger when everything is set up correctly, to avoid error blindness


addons/mod_loader/api/log.gd
```diff
@@ -413,6 +427,9 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 			if verbosity >= VERBOSITY_LEVEL.DEBUG:
 				print(log_entry.get_prefix() + message)
 				_write_to_log_file(log_entry.get_entry())
+		"hint":
+			if OS.has_feature("editor") and verbosity >= VERBOSITY_LEVEL.DEBUG:
+				print_rich("[color=#70bafa]%s[/color]" % (log_entry.get_prefix() + message))
```
KANAjetzt: Can we store the color in a `const` or in `options_profile.gd`?

KANAjetzt: image

For some reason, `[/color]` gets printed. Strangely, removing `[/color]` seems to be fine. 😅

KANAjetzt (approved):
nice 👍  - is the `[/color]` thing not present on 4.2?

addons/mod_loader/api/log.gd
Qubus0: which version did you test on? just 4.3.stable? Does not get printed for me 🤔 not closing the tag leaks color to other messages in 4.1 though

Qubus0: apparently print_rich has an issue with line breaks in bbcode tags. removed the linebreak which also fixes the weird indent
https://github.com/godotengine/godot/issues/102368

Issue #513, opened 2025-01-12
Doc Comment updates


Qubus0:
- [x] Unify the parameter descriptions
  <details>
    <summary>Planned Format</summary>

  ```
  ReturnType method_name(param_name: Type, param_two: Type) modifiers
  Short one line method desc
  
  Parameters:
  - [param param_name] ([Type]): Full param description - what it is and what format it may require
  - [param param_two] ([Type]): ...
  
  Returns: 
  - [code][Type][/code] Description of what the return value is
  or for void: "No return value"
  
  Description:
  Long method description
  
  Examples:
  code with descriptions
  ```
  As a doc comment (some parts are automatically generated and not part of the comment):
  ```gdscript  
  ## Adds all methods from a file as hooks.[br]
  ## [br]
  ## Parameters:[br]
  ## - [param vanilla_script_path] ([String]): File path of the vanilla script to be hooked.[br]
  ## - [param hook_script_path] ([String]): Path to the mod script containing your hook methods.[br]
  ## [br]
  ## No return value[br]
  ## [br]
  ## Description:[br]
  ## The file needs to extend [Object].[br]
  ## The methods in the file need to have the exact same name as the vanilla method
  ## they intend to hook, all mismatches will be ignored.[br]
  ## See: [method add_hook][br]
  ## [br]
  ## Examples:[br]
  ## Addind a script hook file in [code]mod_main.gd[/code][br]
  ## [codeblock]
  ## ModLoaderMod.install_script_hooks(
  ##     "res://tools/utilities.gd",
  ##     extensions_dir_path.path_join("tools/utilities-hook.gd")
  ## )
  ## [/codeblock]
  static func install_script_hooks(vanilla_script_path: String, hook_script_path: String) -> void:
  	pass
  ```
  </details>
- [x] unify bbcode note/callout syntax - must be valid bbcode to look okay in engine and must be unique for us to parse into md

  <details>
    <summary>Planned Format</summary>

    - everything between opening === and closing === is a callout
    - the first line is always discarded in favor of the default titles
    - optional: the type/color can be changed and a custom title can be set
        - by using a (technically invalid but invisible) empty bbcode [color] tag
    - example:
	```gdscript
	## ===[br]
	## [b]Note:[color=note "A note on paths"][/color][/b][br]
	## Your extender script doesn't have to follow the same directory path as the vanilla file,
	## but it's good practice to do so.[br]
	## ===[br]
	```
  </details>

[...]

PR #542, opened 2025-02-08
Doc comment standardization

Qubus0:
works towards 
- #513 

This also adds a project.godot file to the root of the repo, which we need to generate docs via cli.   
this does overlap with the project.godot in /test, so we'll want to think about combining them or not. the only real effect it has is that /test is being ignored when opening the base project in the editor, making manual test runs a bit more tedious


addons/mod_loader/mod_loader.gd
```diff
@@ -214,7 +214,6 @@ func _ready():
 	# Variables initialized with an autoload property cause errors otherwise.
 	if _ModLoaderHooks.any_mod_hooked:
 		if OS.has_feature("editor"):
-			_ModLoaderModHookPacker.start()
```
KANAjetzt, line 217: Why is this removed?

KANAjetzt (approved):
thanks 🤝

addons/mod_loader/mod_loader.gd
Qubus0: that was a testing line that snuck in with one of the last commits. log says no zip will be generated, so no zip should be generated

PR #551, opened 2025-03-20
Add user:// as mod source

### LINKED ISSUE #548: add user:// as mod source
Godot can easily locate the user:// folder on every system no matter where the executable is placed. 

We should only need a new source in the options "user data mods" and then load from user://mods within, just like we are doing from res://mods

DarthMDev:
Fixes #548 
This adds user:// as a mod source option

### DIFF
--- a/addons/mod_loader/mod_loader_store.gd
+++ b/addons/mod_loader/mod_loader_store.gd
@@ -130,6 +130,10 @@ var ml_options := {
 
 	# Array of mods that should be ignored when logging messages (contains mod IDs as strings)
 	ignored_mod_names_in_log = [],
+
+	# If true, ModLoader will load mod ZIPs from the user data directory,
+	# instead of the default location (res://mods)
+	user_data_mods = false,
 }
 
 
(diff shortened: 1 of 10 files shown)


DarthMDev:
If we could get this on the 4.x branch too that be great

addons/mod_loader/mod_loader_store.gd
```diff
@@ -130,6 +130,10 @@ var ml_options := {
 
 	# Array of mods that should be ignored when logging messages (contains mod IDs as strings)
 	ignored_mod_names_in_log = [],
+
+	# If true, ModLoader will load mod ZIPs from the user data directory,
+	# instead of the default location (res://mods)
+	user_data_mods = false,
```
KANAjetzt, line 136: The key im `ml_options` also needs to be added to [options_profile.gd](https://github.com/GodotModding/godot-mod-loader/blob/development/addons/mod_loader/resources/options_profile.gd)

KANAjetzt: You can take a look here 

https://github.com/GodotModding/godot-mod-loader/blob/92cd37ba0136e9d983b6b307900600cfd6705c32/addons/mod_loader/mod_loader_store.gd#L169-L171

to understand how the options are loaded from the resource file.

KANAjetzt (commented):
Thank you for your contribution! Just a quick first look - I will test this later 👍

PR #552, opened 2025-03-22
Add user:// as mod source for 4.x

### LINKED ISSUE #548: add user:// as mod source

DarthMDev:
Fixes #548 (but for 4.x)


KANAjetzt (commented):
Just a quick first look. I recommend adding an override option similar to `override_path_to_mods`.

https://github.com/GodotModding/godot-mod-loader/blob/a165641b5a4644103beb59bc687cf8d0a10a51d7/addons/mod_loader/resources/options_profile.gd#L85

Thanks for your work and let me know if you get stuck on something :)

addons/mod_loader/internal/path.gd
```diff
@@ -291,3 +295,9 @@ static func handle_mod_config_path_deprecation() -> void:
 		ModLoaderLog.error("Failed to rename the config directory with error \"%s\"." % [error_string(error)], LOG_NAME)
 	else:
 		ModLoaderLog.success("Successfully renamed config directory to \"%s\"." % MOD_CONFIG_DIR_PATH, LOG_NAME)
+		
+static func get_path_to_user_mods() -> String:
+	if ModLoaderStore:
+		if ModLoaderStore.ml_options.override_path_to_user_data_mods:
```
Qubus0: could be an `and` here, gdscript short circuits conditions so if the first fails the second won't be checked

addons/mod_loader/resources/options_profile.gd
```diff
@@ -117,3 +117,7 @@ var custom_game_version_validation_callable: Callable
 
 ## Stores the instance of the script specified in [member customize_script_path].
 var customize_script_instance: RefCounted
+
+@export var user_data_mods: bool = true
```
Qubus0: this belongs up at line 76 with the other `load_from_x` options, preferably named `load_from_user_data_mods`
```suggestion
## Indicates whether to load mods from the "mods" folder located in the game's user data directory (user://)
@export var load_from_user_data_mods: bool = true
```

addons/mod_loader/resources/options_profile.gd
```diff
@@ -117,3 +117,7 @@ var custom_game_version_validation_callable: Callable
...
+
+@export_dir var override_path_to_user_data_mods := ""
```
Qubus0: similarly this belongs on line 86. the position kinda matters for the order and collapsible group in the right side inspector.

```suggestion
## Alternative path to a folder containing mods (originally within user://). [br]
## Mod zips should be directly in this folder
@export_dir var override_path_to_user_data_mods := ""
```

Qubus0 (commented):
nice, thank you for your contribution 😄 a few naming and docs things, but looks good otherwise. will need to test at a later date

test/test_options/customize_script/custom_validation.tres
```diff
@@ -24,3 +24,4 @@ override_hook_pack_name = ""
 restart_notification_scene_path = "res://addons/mod_loader/restart_notification.tscn"
 disable_restart = false
 game_version_validation = 2
+override_path_to_user_data_mods
```
Qubus0: was this added automatically by godot or manually? wondering why the `= ""` is missing compared to the other .tres

DarthMDev: my bad must've been a mistake

PR #554, opened 2025-04-20
Add check to see if GodotPCKExplorer executable is present before attempting injection

PolygonError:
Was having a hard time figuring out why the Mod Loader setup script wasn't creating a copy of the main pck file, the reason was because the script just assumes the GodotPCKExplorer executable exists.
If it doesn't exist, it just assumes it worked and continues on.

This pull request just adds a check to see if the executable exists, I tried to add this under handle_injection(), but was having issues with it not exiting properly.
Please make changes to the wording of the error or the location of the check if needed.


PolygonError:
sorry for all the fix up commits, i'm hopeless with git

PR #533, opened 2025-01-30
refactor: use GDRE for self setup

KANAjetzt:
> [!IMPORTANT]
> Currently, self-setup (and Mod Hooks in general) only works if the game exports GDScript in plain text. Before Godot 4.3, this is always the case.  

> [!NOTE]
> Injection is currently only supported for Windows. 

This PR replaces [PckExplorer](https://github.com/DmitriySalnikov/GodotPCKExplorer) with [GDRE](https://github.com/GDRETools/gdsdecomp).  

## Tested With  
- Buckshot Roulette (4.1)  
- RAM Random Access Mayhem Demo (4.2.1)  
- Nodebuster (4.2.2)  
- My test game [Flashit](https://kanamedia.itch.io/flashit) (4.3) 
- DEEP SPACE CACHE (4.3)

## Downloads
Download the zip file for your operating system and unpack the `addons` directory next to the game executable.

[ModLoader-Self-Setup_7.0.0-WIN.zip](https://github.com/user-attachments/files/18814947/ModLoader-Self-Setup_7.0.0-WIN.zip)

PR #555, opened 2025-04-20
Add check for GodotPCKExplorer executable before attempting injection

PolygonError:
Was having a hard time figuring out why the Mod Loader setup script wasn't creating a copy of the main pck file, the reason was because the script just assumes the GodotPCKExplorer executable exists.
If it doesn't exist, it just assumes it worked and continues on.

This pull request just adds a check to see if the executable exists
Please make changes to the wording of the error or the location of the check if needed.


KANAjetzt:
Hi thanks for your PR - in #533 we switched to GDRE for the self setup.

PR #556, opened 2025-04-20
Fix string placeholders and missing log name argument in mod loader setup script

PolygonError:


KANAjetzt:
thanks for your PR 👍 - this has been fixed in #533

PR #557, opened 2025-04-20
Add Parking Garage Rally Circuit to Games Made Moddable by This Project

MichaelHinrichs:


Qubus0:
While it looks like walaber did change their stance on modding in general which is nice to see, it looks like they are using a custom mod loader in their project, not ours specifically (from skimming through their docs here https://walaber-ent.github.io/pgrc-mod/overview/)

MichaelHinrichs:
It appears I was misinformed then. Should i close this pull request?

KANAjetzt:
We can add it to org page under [Moddable Godot games using different systems](https://github.com/GodotModding#moddable-godot-games-using-different-systems). Glad to see modding support for the game 👍

Issue #450, opened 2024-11-18
prettier logging


Qubus0:
this includes prefixing the source file that started the log

```gdscript
static func log_prefix(stack) -> String:
    if len(stack) > 1:
        var call_site = stack[1]
        var basename = call_site["source"].get_file().get_basename()
        var line_num = str(call_site.get("line", 0))
        if call_site["source"].match("*/test/*"):
            return "{" + basename + ":" + line_num + "}: "
        elif call_site["source"].match("*/addons/*"):
            return "<" + basename + ":" + line_num + ">: "
        else:
            return "[" + basename + ":" + line_num + "]: "
    return ""
```
ref https://github.com/russmatney/log.gd/blob/main/addons/log/log.gd#L516-L527

and also print_rich when in the editor to use
- colors for each log level and 
- reduce visual noise by hiding information like timestamps or 
- the mod loader log names within bbcode hint tags (`[hint={tooltip text displayed on hover}]{text}[/hint]`)

PR #562, opened 2025-06-17
use `rich_print` when in-editor

ZackeryRSmith:
Allows for a much cleaner separation of the logs. This PR introduces scaffolding for the options profile. Similar to how `hint_color` currently is. Works towards https://github.com/GodotModding/godot-mod-loader/issues/450
BEFORE | AFTER
:-:|:-:
  |

### DIFF
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -47,7 +47,6 @@ static var ignored_mods: Array[String] = []
 static var warning_color := Color("#ffde66")
 static var success_color := Color("#5d8c3f")
 static var info_color := Color("#70bafa")
-static var debug_color := Color("#00000000")
 static var hint_color := Color("#b293fa")
 
 ## This Sub-Class represents a log entry in ModLoader.
@@ -450,7 +449,7 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 				_write_to_log_file(log_entry.get_entry())
 		"debug":
 			if verbosity >= VERBOSITY_LEVEL.DEBUG:
-				_print_rich("", log_entry.get_prefix() + message, debug_color)
+				print(log_entry.get_prefix() + message)
 				_write_to_log_file(log_entry.get_entry())
 		"hint":
 			if OS.has_feature("editor") and verbosity >= VERBOSITY_LEVEL.DEBUG:
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -43,7 +43,11 @@ static var verbosity: VERBOSITY_LEVEL = VERBOSITY_LEVEL.DEBUG
 ## Array of mods that should be ignored when logging messages (contains mod IDs as strings)
 static var ignored_mods: Array[String] = []
 
-## Highlighting color for hint type log messages
+# TODO: use options profile here instead of hard coding
+static var warning_color := Color("#ffff00")
+static var success_color := Color("#008000")
+static var info_color := Color("#4169e1")
+static var debug_color := Color("#ffffff")
 static var hint_color := Color("#70bafa")
 
 ## This Sub-Class represents a log entry in ModLoader.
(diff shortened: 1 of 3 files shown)


ZackeryRSmith:
~~As for using the options profile scaffolds. If ModLoaderStore is nil, _which in most cases it isn't_, we could just use fallback colors.~~

~~Alternatively, ModLoaderStore logs get to be bland and colorless. I'm reaching for this option, as these logs are very minimal _(literately just a single line)_ and at the top.~~

Scratch all that, just now seeing `_configure_logger`

ZackeryRSmith:
OLD | NEW
:-:|:-:
 |

KANAjetzt:
image

Two things:
- Do we want DEBUG to be bold as well?
- In the editor, we can remove the ERROR prefix.

ZackeryRSmith:
> Do we want DEBUG to be bold as well?

I'm going to argue, no. Debug logs are, well debug. Color and boldness draws attention. It's common practice to reserve color for status', warnings, or errors. Giving debug logs a color elevates them visually to the same importance, which for me breaks the mental model. Where you can just filter the logs for your mods debug messages, I find personally that keeping debug unbold is the way to go. For reference this is what bold debugs look like

###### Original colored log screenshot in this PR


> In the editor, we can remove the ERROR prefix.

Never noticed that, but you're right, let me make that change

KANAjetzt:
Fair point. I think there is still enough separation between the log types with the color alone, and the bold prefix ties all the Mod Loader logs together.
image

Adding a bold toggle in the options for each log type might be a bit excessive, but maybe just one for debug?

ZackeryRSmith:
Agree with you there. Let's make bold the default option for debug logs with an optional toggle. I'll make the changes once I get home

addons/mod_loader/api/log.gd
```diff
@@ -99,9 +104,15 @@ class ModLoaderLogEntry:
 
 	## Get the prefix string for the log entry, including the log type and mod name.[br]
 	## [br]
+	## [b]Parameters:[/b][br]
+	## [param exclude_type] ([bool]): (Optional) If true, the log type (e.g., DEBUG, WARN) will be excluded from the prefix. Default is false.[br]
+	## [br]
 	## [b]Returns:[/b] [String]
-	func get_prefix() -> String:
-		return "%s %s: " % [type.to_upper(), mod_name]
+	func get_prefix(exclude_type: bool = false) -> String:
```
KANAjetzt, line 111:
```suggestion
	func get_prefix(exclude_type := false) -> String:
```

addons/mod_loader/api/log.gd
```diff
@@ -99,9 +104,15 @@ class ModLoaderLogEntry:
...
+		return "%s%s: " % [
+			"" if exclude_type else type.to_upper() + " ",
```
KANAjetzt, line 113:
```suggestion
			"" if exclude_type else "%s " % type.to_upper(),
```

addons/mod_loader/api/log.gd
```diff
@@ -391,6 +402,17 @@ static func get_all_entries_as_string(log_entries: Array) -> Array:
 # Internal log functions
 # =============================================================================
 
+static func _print_rich(prefix: String, message: String, color: Color, bold := true):
```
KANAjetzt, line 405:
```suggestion
static func _print_rich(prefix: String, message: String, color: Color, bold := true) -> void:
```

addons/mod_loader/api/log.gd
```diff
@@ -391,6 +402,17 @@ static func get_all_entries_as_string(log_entries: Array) -> Array:
...
+	if OS.has_feature("editor"):
```
KANAjetzt, line 406:
```suggestion
	if ModLoaderStore.has_feature.editor:
```

addons/mod_loader/api/log.gd
```diff
@@ -422,25 +444,32 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 			_write_to_log_file(JSON.stringify(get_stack(), "  "))
 			assert(false, message)
 		"error":
-			printerr(log_entry.get_prefix() + message)
+			if OS.has_feature("editor"):
```
KANAjetzt, line 447:
```suggestion
			if ModLoaderStore.has_feature.editor:
```

KANAjetzt (approved):
Only some small nitpicks, looks very nice, ty👍

ZackeryRSmith:
I'll write a follow up PR which addresses the suggested changes

PR #566, opened 2025-06-28
style: resolve type and style requests from #562

ZackeryRSmith:
A tiny PR (+10-10) which applies the changes outlined by @KANAjetzt in #562

Auto merge was turned on for the PR, and the changes of #562 were merged before I could resolve the nit picks. This PR resolves those.

### DIFF
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -468,8 +468,10 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 				_print_rich(log_entry.get_prefix(), message, debug_color, debug_bold)
 				_write_to_log_file(log_entry.get_entry())
 		"hint":
-			if ModLoaderStore.has_feature.editor:
-				if verbosity >= VERBOSITY_LEVEL.DEBUG:
+			if (
+				ModLoaderStore.has_feature.editor and
+				verbosity >= VERBOSITY_LEVEL.DEBUG
+			):
 					_print_rich(log_entry.get_prefix(), message, hint_color)
 
 
--- a/addons/mod_loader/api/log.gd
+++ b/addons/mod_loader/api/log.gd
@@ -403,7 +403,8 @@ static func get_all_entries_as_string(log_entries: Array) -> Array:
 # =============================================================================
 
 static func _print_rich(prefix: String, message: String, color: Color, bold := true) -> void:
-	if ModLoaderStore.has_feature.editor:
+	var in_editor: bool = OS.has_feature("editor") if not ModLoaderStore else ModLoaderStore.has_feature.editor
+	if in_editor:
 		var prefix_text := "[b]%s[/b]" % prefix if bold else prefix
 		print_rich("[color=%s]%s[/color]%s" % [
 			color.to_html(false),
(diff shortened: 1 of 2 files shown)


addons/mod_loader/api/log.gd
```diff
@@ -402,8 +402,9 @@ static func get_all_entries_as_string(log_entries: Array) -> Array:
 # Internal log functions
 # =============================================================================
 
-static func _print_rich(prefix: String, message: String, color: Color, bold := true):
-	if OS.has_feature("editor"):
+static func _print_rich(prefix: String, message: String, color: Color, bold := true) -> void:
+	var in_editor: bool = OS.has_feature("editor") if not ModLoaderStore else ModLoaderStore.has_feature.editor
```
KANAjetzt: Right, log.gd is a tool script. In this case, we can use the simple `OS.has_feature("editor")` check.

We could also add a small util to [addons/mod_loader/internal/godot.gd](https://github.com/GodotModding/godot-mod-loader/blob/4.x-dev/addons/mod_loader/internal/godot.gd).

addons/mod_loader/api/log.gd
```diff
@@ -468,8 +469,9 @@ static func _log(message: String, mod_name: String, log_type: String = "info", o
 				_print_rich(log_entry.get_prefix(), message, debug_color, debug_bold)
 				_write_to_log_file(log_entry.get_entry())
 		"hint":
-			if OS.has_feature("editor") and verbosity >= VERBOSITY_LEVEL.DEBUG:
-				_print_rich(log_entry.get_prefix(), message, hint_color)
+			if ModLoaderStore.has_feature.editor:
+				if verbosity >= VERBOSITY_LEVEL.DEBUG:
```
KANAjetzt:
```suggestion
			if ModLoaderStore.has_feature.editor and verbosity >= VERBOSITY_LEVEL.DEBUG:
```

Does it not work like this?

KANAjetzt (commented):
One other small thing, it is helpful to create a separate feature branch for your PRs, it makes it easier for the reviewers to check the branch out. If the PR is using the dev branch directly, the reviewer will most likely have to delete their dev branch first.

Thanks! 😁

addons/mod_loader/api/log.gd
ZackeryRSmith: It does, I think my code editor automatically did this because it went over the 80 character line limit. Personally I find it a bit easier to read 🙈

ZackeryRSmith:
> One other small thing, it is helpful to create a separate feature branch for your PRs, it makes it easier for the reviewers to check the branch out. If the PR is using the dev branch directly, the reviewer will most likely have to delete their dev branch first.
> 
> Thanks! 😁

Sorry about any issue my PRing shenanigans caused, I'll keep this in mind from now on!

addons/mod_loader/api/log.gd
ZackeryRSmith: > We could also add a small util to [addons/mod_loader/internal/godot.gd](https://github.com/GodotModding/godot-mod-loader/blob/4.x-dev/addons/mod_loader/internal/godot.gd).

Seems a bit too much for a function call which is already simple enough. I'll switch back to using `OS.has_feature("editor")`

addons/mod_loader/api/log.gd
KANAjetzt: Then this is the way to go 😁 

```suggestion
			if (
				ModLoaderStore.has_feature.editor and
				verbosity >= VERBOSITY_LEVEL.DEBUG
			):
```

KANAjetzt (approved):

PR #567, opened 2025-06-28
chore: backport 4.x to 4.0

ZackeryRSmith:
Loses engine hex version checking, but seeing as we're below version 4.2 anyway the check isn't required anymore

### DIFF
--- a/addons/mod_loader/internal/godot.gd
+++ b/addons/mod_loader/internal/godot.gd
@@ -12,9 +12,6 @@ const AUTOLOAD_CONFIG_HELP_MSG := "To configure your autoloads, go to Project > 
 const ENGINE_VERSION_HEX_4_2_2 := 0x040202
 const ENGINE_VERSION_HEX_4_2_0 := 0x040200
 
-static var engine_version_hex: int = Engine.get_version_info().hex
-
-
 # Check autoload positions:
 # Ensure 1st autoload is `ModLoaderStore`, and 2nd is `ModLoader`.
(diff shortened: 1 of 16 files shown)


addons/mod_loader/internal/godot.gd
```diff
@@ -106,11 +103,3 @@ static func get_autoload_index(autoload_name: String) -> int:
 	var autoload_index := autoloads.find(autoload_name)
 
 	return autoload_index
-
-
-static func is_version_below(version_hex: int) -> bool:
```
KANAjetzt, line 111: looks like `is_version_below` is still called in [file.gd](https://github.com/GodotModding/godot-mod-loader/blob/2ebd87f9c38017219b47f93eb7ecda9b3754fbfe/addons/mod_loader/internal/file.gd#L199)

KANAjetzt (approved):
I'm assuming you tested hook gen and mod loading on 4.0? 😁
The code looks good, only one small comment about the version check.
You can add an issue for that, just to keep track of it. 👍

ZackeryRSmith:
> I'm assuming you tested hook gen and mod loading on 4.0? 😁

You might be disappointed, I didn't test hook gen 😢
Mod loading, resource overwriting, and so on does work though. Let me test hook gen right now!

ZackeryRSmith:
Hook gen is broken, but due to my lack of knowledge on the subject I don't even have an idea of where to look...

I'll push my changes currently, where still, everything works other than hook gen.

PR #568, opened 2025-07-22
fix: incorrect paths in new_global_classes

ppebb:
See title. Certain paths are seemingly outdated. Had to edit these myself for the self-install to work.


KANAjetzt (approved):
Thank you!

PR #573, opened 2025-08-28
Mac requires 1 more escaped folder to find the workshop contents

kilian-diener:
Probably same goes for 4.x-dev

### DIFF
--- a/addons/mod_loader/internal/third_party/steam.gd
+++ b/addons/mod_loader/internal/third_party/steam.gd
@@ -67,7 +67,10 @@ static func _get_path_to_workshop() -> String:
 
 	# Traverse up to the steamapps directory (ie. `cd ..\..\` on Windows)
 	var path_array := game_install_directory.split("/")
-	path_array.resize(path_array.size() - 3)
+	var folders_to_escape := 3
+	if OS.get_name() == "OSX":
+		folders_to_escape = 4  # on Mac additional escape from GameName.app is needed
+	path_array.resize(path_array.size() - folders_to_escape)
 
 	# Reconstruct the path, now that it has "common/GameName" removed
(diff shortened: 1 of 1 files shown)


Qubus0:
Which game has this issue? We might need to be more specific than just checking for mac since some steam games don't have the app bundle layer, so this issue does not always exist. We might just have to check both paths

kilian-diener:
I'm currently setting up a Mac version of Brotato. 
The path array is [..., Steam, steamapps, common, Brotato, Brotato.app, ]
With removing 3 entries this results in this workshop path: Steam/steamapps/common/workshop/content/1942280
With removing 4 it resolves correctly to: Steam/steamapps/workshop/content/1942280

I built the game as a ZIP, unpack it and upload the resulting Brotato.app. Maybe it would be different for a .dmg exported game.

Qubus0:
ah we accounted for that in 4.x with this .app check

https://github.com/GodotModding/godot-mod-loader/blob/f2d9954bd72a2957b1531a89e196bd87fe21d29e/addons/mod_loader/internal/path.gd#L22-L25

which is not present in 3.x 

https://github.com/GodotModding/godot-mod-loader/blob/62a08119b217899144feec7ae99571428052473a/addons/mod_loader/internal/path.gd#L17-L19

also hype to hear about brotato for mac 👀

kilian-diener:
Ah yeah that works as well. However, this may break setups for people using the mods folder without workshop since now the mods folder is expected to be outside the .app folder right? But whatever approach you guys prefer. I'll have it fixed for us anyway.

Qubus0:
yep, true. most users don't know you can even look inside the .app bundle, so that works 😄 we'll take care of it, thanks for your contribution 👍

PR #576, opened 2025-10-16
fix: Add missing keys in global_script_class_cache_mod_loader.cfg for 4.5.x

Bernier154:
### **What is this solving?**

I was not able to setup de mod-loader on a 4.5.x game, and found out why.

This line in the engine makes the combined class cache invalid
https://github.com/godotengine/godot/blob/20430236e749306e1968656ffdc7f78e502a0ccc/core/object/script_language.cpp#L295
image

It was introduced in version 4.5.1
https://github.com/godotengine/godot/commit/318af42020cd349c5c9c3e4c88aa87249ca224b0

This PR is the essential modification that I brought to my files to make it work.

I made these change manually, but it work.  Maybe load the extension in the editor and copy the new classes? I added `is _tool` to class that were having `@tool` at the top and no abstract classes so it must me okay.

### **Suggestion**

GDRE tool version needed to also be bumped, i switched it with version 1.0.2 to make it work. 2.0.0beta8 was also working but while trying to debug i tested multiple version. 

The actual version supplied in the release outputted an error saying something like "pack version 3 unsupported" on something like that. ( The current target game was compiled as 4.5 beta2 )


KANAjetzt (approved):
Thanks!

Issue #490, opened 2024-12-14
Mod Config: API Method to duplicate a config


KANAjetzt:
A method to duplicate an existing config (in most cases the default config) would simplify the creation of a modifiable config since the default config is read-only. 

For example:  
`duplicate_config(config: ModConfig, config_name: String) -> ModConfig:`

can be called as:
`ModLoaderConfig.duplicate_config(ModLoaderConfig.get_default_config(MOD_ID), "custom")`

https://github.com/GodotModding/godot-mod-loader/blob/aab3aaa089f75404f07dd790aa193bb430a5582a/addons/mod_loader/api/config.gd#L12-L69

PR #580, opened 2025-10-31
feat: :sparkles: added `duplicate_config`

KANAjetzt:
works towards 
- #490

### DIFF
--- a/addons/mod_loader/api/config.gd
+++ b/addons/mod_loader/api/config.gd
@@ -129,6 +129,18 @@ static func delete_config(config: ModConfig) -> bool:
 	return true
 
 
+# Duplicate an existing ModConfig
+#
+# Parameters:
+# - config (ModConfig): The ModConfig object to be duplicated.
+# - config_name (String): The name of the new configuration.
+#
+# Returns:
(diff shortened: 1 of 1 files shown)


addons/mod_loader/api/config.gd
```diff
@@ -129,6 +129,18 @@ static func delete_config(config: ModConfig) -> bool:
...
+# Duplicate an existing ModConfig
+#
+# Parameters:
+# - config (ModConfig): The ModConfig object to be duplicated.
+# - config_name (String): The name of the new configuration.
+#
+# Returns:
+# - ModConfig: The duplicated ModConfig object if successful, or null otherwise.
+static func duplicate_config(config: ModConfig, config_name: String) -> ModConfig:
+	return create_config(config.mod_id, config_name, config.data)
```
Qubus0: Passing the .data in directly might create issues since both dictionaries would be the same reference, so changing the original will change the dupe. I think .data.duplicate(true) is what we want

KANAjetzt: Beginner mistake 😄 ty

```suggestion
	return create_config(config.mod_id, config_name, config.data.duplicate(true))
```

Qubus0 (approved):

PR #587, opened 2026-02-01
fix: detect indentation in source for generating hook code

### LINKED ISSUE #575: hook preprocessing spaces vs tabs
we assume tabs for indentation, but that breaks the script when hooking a vanilla file that uses spaces.

detecting which spacing is used might be the difficult part, though we could also pass it as a parameter

NanobotZ:
Added detecting indentation for generating hook code.
Fixes cases when script's source doesn't use single tabs.
Fixes #575

### DIFF
--- a/addons/mod_loader/internal/mod_hook_preprocessor.gd
+++ b/addons/mod_loader/internal/mod_hook_preprocessor.gd
@@ -63,6 +63,7 @@ func process_script(path: String, enable_hook_check := false, method_mask: Array
 	var current_script := load(path) as GDScript
 	var source_code := current_script.source_code
 	var source_code_additions := ""
+	var detected_indentation := _detect_indentation(source_code)
 
 	# We need to stop all vanilla methods from forming inheritance chains,
 	# since the generated methods will fulfill inheritance requirements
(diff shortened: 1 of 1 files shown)


kyfex-uwu:
works for me with the game Ultrapool (uses 4 spaces instead of tabs)

KANAjetzt (approved):
thanks!

KANAjetzt:
We should add [an option](https://github.com/GodotModding/godot-mod-loader/blob/4.x-dev/addons/mod_loader/resources/options_profile.gd) that allows developers to set this when integrating the loader, so detection can be skipped if the option is set.

Ech0klang:
Manually merged this with PR #588 and found it NOT working for a game in a custom version of godot 4.6.
The decompiled version of a script has 5 spaces as indentation, without modification this produced indentation of 1 tab and 5 spaces. I've temporarily "fixed" this for me by overwriting the "Default state" (as set at the end of the modified script)

NanobotZ:
> Manually merged this with PR #588 and found it NOT working for a game in a custom version of godot 4.6. The decompiled version of a script has 5 spaces as indentation, without modification this produced indentation of 1 tab and 5 spaces. I've temporarily "fixed" this for me by overwriting the "Default state" (as set at the end of the modified script)

Could you provide some examples?
I've tested it with some random script text having 5 spaces as indentation, and detected indentation was 5 spaces.

Issue #535, opened 2025-01-31
[4.x] Mod Hooks: Binary tokens


KANAjetzt:
Mod Hooks only work if the source code is readable.  
With Godot 4.3, binary tokenization has been reintroduced and is the default export setting.  

![Image](https://github.com/user-attachments/assets/5a4d0a3e-4986-4131-8c23-7d70f044883e)  

- https://github.com/godotengine/godot/pull/87634  

For game developers integrating the Mod Loader, adding a step in the documentation to switch to text mode will be sufficient.  

![Image](https://github.com/user-attachments/assets/628487f5-f08d-4e66-acf0-149279276410)  

For self-setup, we will need to use GDRE to decompile scripts that require a hook.  

Script extensions are not affected by this.

PR #588, opened 2026-02-01
feat: decompile and inject needed scripts before creating hooks

### LINKED ISSUE #586: Decompile scripts when hooking a game that has compressed scripts
Currently the hook preprocessor simply tries loading a script provided by the path, but that approach doesn't work if the game was exported with compressed scripts (.gdc).

I did some tests and it is possible to decompile the compressed .gdc scripts and patch them into the .pck - adding them as .gd files (not replacing the .gdc).
Doing that makes the mod loader properly generate hooks when launching the game.
It is important to also patch the *.gd.remap files then doing that, otherwise the game will still load the compiled .gdc scripts.

My proof of concept:
```python
from pathlib import Path
import os
import shutil
import subprocess
import sys

pck_path = 'Game.pck'
temp_folder = 'temp'
tool_path = 'gdre_tools.exe'


temp_path = Path(temp_folder)

def recover_scripts_and_remaps():
    # recover/decompile all scripts, we do this first because it cleans the temp folder
    subprocess.call([tool_path, '--headless', f'--recover={pck_path}', '--include=res://**/*.gdc', f'--output={temp_folder}'])
    # extract all the .gd.remap files
    subprocess.call([tool_path, '--headless', f'--extract={pck_path}', '--include=res://**/*.gd.remap', f'--output={temp_folder}'])

def fix_remaps():
    # go through all .gd.remap files and make them point back to the .gd file
    for remap_file_path in temp_path.rglob('*.gd.remap'):
        content = ''
        with open(remap_file_path, 'rt') as remap_file:
            content = remap_file.read()
            content = content.replace('.gdc"', '.gd"')
        with open(remap_file_path, 'wt') as remap_file:
            remap_file.write(content)

def get_max_cmd_length():
    if os.name == 'nt':
        return 30000
    return 100000

MAX_CMD_LENGTH = get_max_cmd_length()

def collect_patch_args(patterns):
    patch_args = []
    seen_files = set()

    for pattern in patterns:
        for file_path in temp_path.rglob(pattern):
            if file_path in seen_files:
                continue
            seen_files.add(file_path)

            source_rel = file_path.relative_to(temp_path)
            source_path = str(temp_path / source_rel)
            dest_path = 'res://' + source_rel.as_posix()

            patch_args.append(f'--patch-file={source_path}={dest_path}')

    return patch_args


def chunk_args(base_args, patch_args):
    chunks = []
    current = []

    base_len = sum(len(a) + 1 for a in base_args)
    current_len = base_len

    for arg in patch_args:
        arg_len = len(arg) + 1

        if current_len + arg_len > MAX_CMD_LENGTH:
            chunks.append(current)
            current = [arg]
            current_len = base_len + arg_len
        else:
            current.append(arg)
            current_len += arg_len

    if current:
        chunks.append(current)

    return chunks


def patch_many(patterns):
    patch_args = collect_patch_args(patterns)

    if not patch_args:
        return

    base_args = [
        tool_path,
        '--headless',
        f'--pck-patch={pck_path}',
        f'--output={pck_path}',
    ]
    
    # chunking the --patch-file commands to avoid reloading GDRE each file
    chunks = chunk_args(base_args, patch_args)

    for i, chunk in enumerate(chunks, 1):
        args = base_args + chunk
        print(f'Running patch batch {i}/{len(chunks)} with {len(chunk)} files')
        subprocess.check_call(args)


recover_scripts_and_remaps()
fix_remaps()
patch_many(['*.gd', '*.gd.remap'])

shutil.rmtree(temp_path)
```

In theory it might be possible to do this without patching the game's PCK, but I haven't tested that:
- while preparing the hooks, check whether there is a compressed script, if it's not then continue processing the current way,
- decompile with GDRE to .gd file, apply patches onto that, put it into the zip,
- also extract .gd.remap file and fix the path inside from .gdc to .gd (or generate a new .gd.remap file), also put into the zip.

NanobotZ:
Adds automatic script decompilation using GDRE, so that ModHookPreprocessor can properly load and analyze the scripts and generate valid hook code.

If game contains compressed/compiled scripts (checking for .gdc files), it works by:
- decompiling needed scripts,
- extracting their .remap files and fixing them (pointing from .gdc file back to .gd),
- putting both into a zip,
- loading the zip immediately as a resource pack (please let me know if there's an easier way for loading a script as a GDScript resource).

After hook processing, modified source together with the .remap file are packed into the mod-hooks.zip, making the game load the source properly.

I tested this using only one game (Coal LLC) with two scripts in a work-in-progress mod, additional testing might be required.

Fixes #586 
Works towards #535

### DIFF
--- a/addons/mod_loader/internal/mod_hook_packer.gd
+++ b/addons/mod_loader/internal/mod_hook_packer.gd
@@ -7,6 +7,7 @@ extends RefCounted
 
 const LOG_NAME := "ModLoader:ModHookPacker"
 
+static var ModLoaderSetup: Object = load("res://addons/mod_loader/mod_loader_setup.gd")
 static var ModLoaderSetupLog: Object = load("res://addons/mod_loader/setup/setup_log.gd")
 static var ModLoaderSetupUtils: Object = load("res://addons/mod_loader/setup/setup_utils.gd")
 
@@ -108,7 +109,7 @@ static func _decompile_and_load_if_needed() -> bool:
 	var game_base_dir: String = ModLoaderSetupUtils.get_local_folder_dir()
 	# C:/path/to/game/addons/mod_loader
 	var mod_loader_dir: String = game_base_dir + "addons/mod_loader/"
-	var gdre_path := mod_loader_dir + "vendor/GDRE/gdre_tools.exe"
+	var gdre_path: String = mod_loader_dir + ModLoaderSetup.get_gdre_path()
 	var decomp_dir := mod_loader_dir + "decomp/"
 	var zip_path := decomp_dir + "inject.zip"
 
--- a/addons/mod_loader/internal/mod_hook_packer.gd
+++ b/addons/mod_loader/internal/mod_hook_packer.gd
@@ -7,8 +7,12 @@ extends RefCounted
 
 const LOG_NAME := "ModLoader:ModHookPacker"
 
+static var ModLoaderSetupLog: Object = load("res://addons/mod_loader/setup/setup_log.gd")
+static var ModLoaderSetupUtils: Object = load("res://addons/mod_loader/setup/setup_utils.gd")
 
 static func start() -> void:
+	_decompile_and_load_if_needed()
+
 	ModLoaderLog.info("Generating mod hooks .zip", LOG_NAME)
 	var hook_pre_processor = _ModLoaderModHookPreProcessor.new()
 	hook_pre_processor.process_begin()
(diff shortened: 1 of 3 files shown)


kyfex-uwu:
works for me, hooked into compiled files in Ultrapool

KANAjetzt (commented):
Thanks. Great work!

I feel like it would be nice to decouple this into `self setup` and `loader integrated`. Because we only want to use GDRE in a self setup scenario.

KANAjetzt (approved):
Created 

- #604 

to keep track of the decoupling.

PR #600, opened 2026-07-02
Don't log error when a mod folder isn't found

kilian-diener:
Per default, the mod loader throws errors when the mods-unpacked folder or the Steam workshop folder isn't found.
However, on a clean install of a game without any mods installed, this is the expected setup.
Only once you download something from the Steam workshop or load a mod from some other source do these folders exist.
Now instead of reporting this configuration as an ERROR, its only an INFO resulting in a clean startup.

Before:
Screenshot_20260702_163839
After:
image


KANAjetzt (approved):
thanks!

